PR #19, opened 2021-08-01
front: show diff and logs

### LINKED ISSUE #9: Integrate diff.py

### LINKED ISSUE #17: Add error pane to scratch view
The backend now puts stderr from the compiler into an "errors" field in the compile response object. We should add a read-only panel to the scratch view to show the user these errors.

nanaian:
image
resolves #9 
resolves #17


nanaian:
resolves #9 
resolves #17

PR #30, opened 2021-08-03
Adding docker-compose.yaml

mkst:


ethteck:
Looks great! The only thing I see we should maybe change is to name the entrypoint shell scripts docker_entrypoint so it's clear they're docker-related

PR #35, opened 2021-08-04
use sqlite by default

### LINKED ISSUE #28: Configure sqlite as the development db
It's probably deadly simple

nanaian:
Configures the default `.env` to use SQLite by default, meaning a postgresql server is no longer required.

Also removes postgres from requirements.txt, as its only needed in production (and even then its a choice on behalf of whoever's hosting as to which db they use).

Resolves #28.


backend/requirements.txt
```diff
@@ -1,5 +1,4 @@
 Django
-psycopg2
```
ethteck, line 2: this is still a production requirement, so I think we should probably make a requirements-prod file or something if you wanna remove it from here

PR #46, opened 2021-08-06
restyle whole site

nanaian:
image
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ build
 backend/local_files/
 postgres/
 *.db
+.DS_Store
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -6,6 +6,7 @@
     },
     "dependencies": {
         "@monaco-editor/react": "^4.2.1",
+        "@primer/octicons-react": "^15.0.0",
         "preact": "^10.5.13",
         "preact-router": "^3.2.1",
         "react-hot-toast": "^2.1.0",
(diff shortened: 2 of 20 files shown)


ethteck:
Thanks for doing this - I think it's a huge improvement across the board. I just tested it out and have a few notes:

The compiler dropdown doesn't really look like a dropdown anymore (it doesn't appear obvious that you can change it). I like the way it looks, but maybe some indicator on the right side that it can be clicked or expanded would be helpful.

Only the very right edge of the diff/code separator bar seems draggable now, where it used to be the whole thing.

And lastly, there seems to be some bug regarding when a new compile request is sent through. I can enter text and wait ages without it doing one, and then I have to trigger it manually. It appears a compile call is going out right after any inital keystroke, and then subsequent keystrokes are not triggering one, even after a delay. I'd suggest we wait until some time has passed before the last keystroke before sending out a call, such as 500 ms or something. That and if we're not waiting for a compile request to return and we know the contents of the buffer don't match the diff, we should probably do like a sync-up compile request after some delay. Idk, not sure exactly what's going on there...

ethteck:
Oh, and one last thing: on the index page, the "create a scratch" text should probably have some left padding since it's pushed right up against the window kind of awkwardly. We'll need to design a proper homepage at some point as well but that's for later

nanaian:
The compile thing should be fixed now - could you test it @ethteck?

PR #54, opened 2021-08-09
Move object files from the filesystem into the database

zbanks:
- `object` --> `elf_object`
   - `elf_object` is now a `BinaryField`, to store the `.o` data directly in the DB
   - Added a (bidirectional) migration
   - Renamed to avoid clashing with Python reserved word (happy to change it to anything else!)
- Updated all the wrappers to use `NamedTemporaryFile` to pass things to/from subprocesses
- Deleted `CompilerWrapper.run_objdump()` because it was unused (only the `asm_diff_wrapper.py` version is used right now, and that one seems more generic?)
- Added `psycopg2` to `requirements.txt` & sorted the file (needed to bring up Docker image)

The migration will import the existing data from the filesystem into the DB (or *export* if run in reverse). It doesn't delete any data from the filesystem, because that is tricky to do correctly without risking data loss. (You'd want to wait until after the DB transaction succeeds.) -- and also it feels kind of risky anyways! 

I'd still hope we can do some scheduled maintenance for this migration or something?

Tested locally by creating a scratch, recompiling it, saving it. Tested migrating both forwards & backwards repeatedly.


backend/coreapp/migrations/0021_rename_assembly_asm.py
```diff
@@ -4,6 +4,7 @@
 
 
 class Migration(migrations.Migration):
+    atomic = False
```
zbanks, line 7: This is maybe a tiny bit less safe than it could be on Postgres, but this migration can't run against a sqlite DB without it? And I don't think it'll be an issue moving forward?

backend/coreapp/compiler_wrapper.py
```diff
@@ -34,14 +29,13 @@ def run_compiler(compile_cmd: str, input_path: Path, output_path: Path, compiler
...
             .replace("$CC_FLAGS", cc_flags)
-        compile_command = f"bash -c \"{compile_command}\""
 
         logger.debug(f"Compiling: {compile_command}")
 
         return_code = 0
 
         try:
-            result = subprocess.run(compile_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
+            result = subprocess.run(["bash", "-c", compile_command], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
zbanks, line 38: This removes a shell layer (`shell=True` is basically `bash -c ...`), and helps avoid some escaping issues in `compile_command`.

backend/coreapp/migrations/0021_rename_assembly_asm.py
ethteck: yeah, no problem. I think we'll compress all the migrations soon anyway

backend/coreapp/compiler_wrapper.py
ethteck: great!

nanaian (approved):
🚀

PR #56, opened 2021-08-09
Compiler Option Customisation, Forking, Caching, Bug fixes

### LINKED ISSUE #16: Client/server compilation caching
Given a compiler configuration, c context, c code, and target asm (in the case of a slug-less compile request), cache the compilation result on the server and/or the client so we don't have to invoke the compiler every time this same request is recieved. 

I imagine for things like trying and undoing small changes over and over, this could really dramatically improve the UX of the app. Not to mention that anyone loading a scratch page will get a near-instant result on the screen immediately, in the case of server-side caching.

### LINKED ISSUE #26: Scratch forking
Allow users to create forks from scratches 

- [ ] Add the fork button to the scratch view and allow anyone to click it to create a fork
- [ ] ctrl+s forks for non-owners? (I wanna ask people if this inconsistency with hotkey behavior will be confusing - probably fine though)
- [ ] Store the parent scratch into the new scratch and provide a link to the scratch's parent on the front-end

### LINKED ISSUE #27: Move compiler configuration into the filesystem?
Or at least provide some mechanism for loading compiler configs into the database

### LINKED ISSUE #36: Compiler flag customisation
Provide some pop-up interface for interactively configuring supported compiler flags, rather than just picking from a dropdown list.

e.g. select GCC or IDO, then an -O level, a -g level, and any -f flags.

Related to #27.

### LINKED ISSUE #39: Avoid sending the context again if it hasn't changed from the database's context
Sometimes the context is 200kb or more, so we should really avoid sending the context to the backend if it hasn't changed.

We need to make sure the backend's version matches the front end's context in order to bypass sending it, so my idea is to have the backend to hash the context and give the hash to the front end. The front end will hash it whenever it needs to make a compile request. If the hashes match, it'll send a signal for the context instead like `unchanged`. When the scratch is saved, the backend can return the updated hash so the client always knows what's up.

### LINKED ISSUE #40: Add an "original context" field to Scratch
The database should have a notion of "original context", which is just a copy of the originally submitted context that never changes. We'll need to be able to eventually diff against it to determine changes that were needed in the context for matching.

### LINKED ISSUE #43: Give m2c the context when making the stab

### LINKED ISSUE #47: Support for other arches
Add arch field to Compiler so we can begin to support other arches. We'll be able to use this when invoking asm-differ to get diffing for other arches as well

ethteck:
Fixes #47, fixes #43, fixes #40, fixes #39, fixes #36, fixes #27, fixes #26, fixes #16, among possibly others


ethteck:
Fixes #47, #43, #40, #39, #36, #27, #26, #16

ethteck:
Fixes #43 
Fixes #40 
Fixes #39 
Fixes #36 
Fixes #27 
Fixes #16

ethteck:
Fixes #43

ethteck:
Fixes #40

ethteck:
Fixes #39

ethteck:
Fixes #43 , fixes #40 , fixes #39 , fixes #36 , fixes #27 , fixes #26 , fixes #16

PR #65, opened 2021-08-11
Run compiler commands in an nsjail

zbanks:
- Did not commit any changes to the migrations/models (I had to wipe them & start fresh to test this locally)
- Updated IDO recomp binaries. I built these locally using the latest ido-static-recomp that has my fix to use `getenv()`. 
    - This & the new directory structure makes it much easier to run inside the `nsjail`.
- Renamed `COMPILER_PATH` -> `COMPILER_DIR` to avoid conflicting with [GCC env vars](https://gcc.gnu.org/onlinedocs/gcc/Environment-Variables.html)
- Created a `Sandbox` context manager for handling temporary files & sandboxing.
    - At its simplest, it is a wrapper around `TemporaryDirectory` that returns a `Path` instead of a `str` and honors the `SANDBOX_TMP_PATH` setting.
    - The `Sandbox.run_subprocess(...)` is a replacement for `subprocess.run(...)` that runs the command in a jail (and optionally, in a shell)
    - The `USE_SANDBOX_JAIL` setting can be set to `False` for simpler local development.
    - `sandbox_wrapper.py` is formatted with `black` and passes `mypy` type checks.
- [x] I would like some advice on how to improve the Dockerfile. What's there now for running the `nsjail` binary is basically a placeholder. :grin:

Some security notes:
- I think I needed to enable user namespaces on my machine (this may be enabled by default, depending on the distro) 
    - `echo 'kernel.unprivileged_userns_clone=1' | sudo tee -a /etc/sysctl.d/00-local-userns.conf && sudo service procps restart`
    - [x] If still necessary, add to the README
- We disable seccomp/apparmor at the Docker level: I think this is fine, but in the future someone could write custom seccomp/apparmor profiles for this.
- We use a dedicated tmpfs separate from the container's main `/tmp` so that a malicious jailed process can't use up all of the container's tmp space and cause "big" issues.
    - If multiple jailed processes are running at the same time, they still can't read each others files.
    - But, 1 process could "DoS" the others by using all the tmp space.
- [x] `nsjail` has some good defaults, and I think I enabled the right set of options -- but I'd appreciate a second opinion.
 

Tested the following:
- Created a new scratch
- Compiled a scratch with GCC, IDO 5.3, & IDO 7.1 with multiple command line arguments (with & without errors)
- Generated the diff (uses the `objdump` subprocess)


backend/decompme/settings.py
```diff
@@ -138,5 +138,7 @@
 
 COMPILER_BASE_PATH = BASE_DIR / "compilers"
 LOCAL_FILE_DIR = BASE_DIR / "local_files"
-ASM_OBJECTS_PATH = LOCAL_FILE_DIR / "assemblies"
-COMPILATION_OBJECTS_PATH = LOCAL_FILE_DIR / "compilations"
+
+USE_SANDBOX = True
```
ethteck, line 142: I think we should use the django environ method of setting this, so having `env('USE_SANDBOX')` as the value here and adding a line for it with the default value being True in the object on line 10. And then in the .dev in the repo, we can also specify it as True which makes it easy to disable locally (just change the .env file like you do for other settings)

PR #76, opened 2021-08-23
Arch refactor

### LINKED ISSUE #45: Prevent reorderings when calling modern as
we can probably prepend `.set noreorder` to target asm before building it, or maybe there's an assembler flag

ethteck:
Closes #45


nanaian (approved):
All the frontend stuff looks good! There's definitely some deduplication w.r.t. ArchButton vs CompilerButton but I can look into that later

PR #79, opened 2021-08-25
Support permuter import

simonlindholm:


backend/coreapp/compiler_wrapper.py
```diff
@@ -56,6 +60,32 @@ class CompilerWrapper:
...
         return settings.COMPILER_BASE_PATH
 
+    @staticmethod
+    def arch_from_compiler(compiler: str) -> Optional[str]:
+        cfg = compilers.get(compiler)
+        return cfg["arch"] if cfg else None
+
+    def cc_opts_from_command(compiler: str, compile_command: str) -> str:
+        cfg = compilers[compiler]
+        # TODO: use cfg for this?
```
nanaian, line 70: This is weird, ideally we'd grab every flag given to cc

backend/coreapp/views.py
```diff
@@ -13,18 +13,14 @@
 from .models import Profile, Asm, Scratch
 from coreapp.models import gen_scratch_id
 
+
 def get_db_asm(request_asm) -> Asm:
     h = hashlib.sha256(request_asm.encode()).hexdigest()
+    asm, _ = Asm.objects.get_or_create(hash=h, defaults={
```
nanaian, line 19: Should probably handle the error

backend/coreapp/compiler_wrapper.py
simonlindholm: we must avoid grabbing at least `-o`, `$INPUT`/`$OUTPUT`, and the compiler command itself. Some of them don't make sense to grab, like `-G0`/`-non_shared` which are already part of the base compiler command, and `-I...`/` `-woff 123`, which are fairly harmless but bloats the compiler option UI. A list of allowed options, which we can get from the cfg's setting view, seemed like a reasonable option. Could also go the other way and try to filter out unwanted things, though, just a tad more work.

backend/coreapp/views.py
simonlindholm: not an error, just a boolean saying whether it already exists

backend/coreapp/compiler_wrapper.py
nanaian: Yeah, perhaps we could grab all `-W`, `-f`, `-g`, and `-O` flags?

ethteck (approved):
Once we figure out what to do with the `interesting_flags` discussion, I'm cool with merging this. Thanks Simon for also making the code much cleaner in some places :)

backend/coreapp/compiler_wrapper.py
ethteck: I'm fine with it as-is for now, but yeah I guess having it be part of the cfg is good too (though repeating stuff in each compiler cfg could get annoying)

PR #107, opened 2021-08-27
Add GitHub sign-in & add lint actions

### LINKED ISSUE #83: GitHub integration
- [ ] Log in with GitHub
- [ ] Show username/avatar of fork owners if they have GitHub connected

nanaian:
Fixes #83.


backend/Dockerfile
```diff
@@ -41,3 +41,5 @@ COPY --from=nsjail /bin/nsjail /bin/nsjail
 WORKDIR /backend
 
 ENTRYPOINT ["/backend/docker_entrypoint.sh"]
+
+# TODO: nginx server to proxy-pass frontend/backend in order to preserve cookies
```
nanaian, line 45: fwiw im unsure if this is required

backend/coreapp/views.py
```diff
@@ -201,3 +216,91 @@ def fork(request, slug):
...
+            profile.save()
+            request.session["profile"] = profile.id
+        elif error == "bad_verification_code":
+            return Response({
+                "error": "Invalid or expired GitHub OAuth verification code",
+            }, status=status.HTTP_401_UNAUTHORIZED)
+        else:
+            raise Exception(f"Unknown GitHub login error: {error} - {response['error_description']}")
+
+    else:
```
ethteck, line 291: Since this is far from the original if and it's not completely obvious that this has to be a GET, I think it's worth doing elif GET here

ethteck (commented):
I don't feel super qualified to review the backend stuff, or at least I'd rather have someone with a bit more experience look at that as well. Ideally we can clean up the scratch serializer stuff

backend/coreapp/views.py
nanaian: :+1:

backend/coreapp/views.py
```diff
@@ -201,3 +216,91 @@ def fork(request, slug):
...
     new_scratch.save()
     return Response(ScratchSerializer(new_scratch).data, status=status.HTTP_201_CREATED)
+
+@api_view(["GET", "POST"])
+def user_current(request, slug=None):
+    """
+    Get the logged-in user, or sign in with GitHub
+    """
+
+    profile = Profile.objects.filter(id=request.session.get("profile", None)).first()
```
zbanks, line 226: I think `request.session.get("profile")` is equivalent to `.get("profile", None)`?

Another option is `Profile.objects.get(id=...)` (which will throw `Profile.DoesNotExist` if it doesn't exist). 

(If `"profile"` is not None and the `Profile` doesn't exist, it might even be worthy of raising an error?)

zbanks (commented):
Overall looks good! I only looked at the backend components. 
Just some minor suggestions

backend/coreapp/views.py
simonlindholm: Split GET/POST handlers into separate functions? I've wanted to do that for the `/api/scratch` route.

ethteck: Yeah that makes sense actually, particularly for /api/scratch

nanaian: There's a Django view class specifically for this, I think, so I'll use that

.github/workflows/reviewdog.yml
```diff
@@ -0,0 +1,18 @@
+name: reviewdog
+on:
+  push:
+    branches:
+      - main
+  pull_request:
```
mkst, line 6: I *think* this should be `pull_request_target` such that this action can run on PRs from forks of the repo. If that's what you want :)

backend/coreapp/serializers.py
```diff
@@ -1,11 +1,41 @@
...
+        return {
+            "is_you": user == request.user,
+            "is_anonymous": False,
+            "id": user.id,
+            "username": user.username,
+            "email": user.email,
+            "name": github.details().name if github else user.username,
+            "avatar_url": github.details().avatar_url if github else None,
+            "github_api_url": github.details().url if github else None,
+            "github_html_url": github.details().html_url if github else None,
```
zbanks, line 28: May want to do something like `github_details = github.details() if github else None`? ( in case caching is disabled, e.g. `API_CACHE_TIMEOUT = 0`)

backend/coreapp/views.py
```diff
@@ -215,92 +217,42 @@ def fork(request, slug):
...
         else:
-            raise Exception(f"Unknown GitHub login error: {error} - {response['error_description']}")
+            logout(request)
 
-    else:
-        profile.load_fields_from_github()
-        profile.save()
+            profile = Profile()
+            profile.save()
+            request.profile = profile
```
zbanks, line 246: I think you also need to store the profile id in the `request.session`? I think as-is, the next request will *also* create a new session?

(Could you return an empty response instead? Or a "logout successful" kind of message?)

backend/decompme/settings.py
```diff
@@ -49,8 +54,10 @@
     'django.contrib.sessions.middleware.SessionMiddleware',
     'corsheaders.middleware.CorsMiddleware',
     'django.middleware.common.CommonMiddleware',
-    'django.middleware.csrf.CsrfViewMiddleware',
+    #'django.middleware.csrf.CsrfViewMiddleware',
+    'coreapp.middleware.disable_csrf',
```
zbanks, line 58: I'd still be hesitant about disabling CSRF - but I don't feel strongly enough to hold up this review over it. (My web security knowledge is a few years out-of-date, so I'm leaning more conservative re: security)

backend/coreapp/github.py
```diff
@@ -0,0 +1,111 @@
...
+        related_name="github",
+    )
+    github_id = models.PositiveIntegerField(unique=True, editable=False)
+    access_token = models.CharField(max_length=100)
+
+    class Meta:
+        verbose_name = "GitHub user"
+        verbose_name_plural = "GitHub users"
+
+    def details(self, use_cache: bool = True) -> NamedUser:
```
zbanks, line 45: I don't think the `use_cache` parameter is ever used? (is there a `.details(use_cache=False)` missing somewhere?)

backend/coreapp/github.py
```diff
@@ -0,0 +1,111 @@
...
+        if error == "bad_verification_code":
+            raise BadOAuthCode()
+        elif error:
+            raise Exception(f"GitHub login sent unknown error '{error}'")
+
+        scopes = str(response["scope"]).split(",")
+        if not "public_repo" in scopes:
+            raise MissingOAuthScope("public_repo")
+
+        access_token: str = response["access_token"]
```
zbanks, line 83: May want to unpack `response` first, and raise a specific error for a malformed `response`.

```python
try:
  scope_str = response["scope"]
  access_token = response["access_token"]
except KeyError:
  raise ...
```

(I would also recommend replacing the generic `Exception` on line 77?)

backend/coreapp/middleware.py
```diff
@@ -0,0 +1,68 @@
...
+        return get_response(request)
+
+    return middleware
+
+def set_user_profile(get_response):
+    """
+    Makes sure that `request.profile` is always available, even for anonymous users.
+    """
+
+    def middleware(request: HttpRequest):
```
zbanks, line 32: I think if you change this `HttpRequest` to `Request`, then you can use plain `request.profile = profile` at the bottom of this fn instead of `setattr(...)`

backend/coreapp/middleware.py
```diff
@@ -0,0 +1,68 @@
...
+        if not request.user.is_anonymous:
+            try:
+                profile = request.user.profile
+            except User.profile.RelatedObjectDoesNotExist: # type: ignore
+                pass
+
+        if not profile:
+            id = request.session.get("anonymous_profile_id")
+            if not isinstance(id, int):
+                id = -1
```
zbanks, line 44: `.get(..., -1)`? (+ an `assert isinstance(id, int)` if mypy is complaining?)

Though, I would recommend not using `-1` as a sentinel value -- because this logic would fall apart if there were a Profile in the DB with id `-1`. Can this be refactored to avoid trying `Profile.objects.get(...)` if the anon id isn't in `request.session`?

zbanks (commented):
Again, I only looked at the Python sections. Mostly minor comments, overall looking good (and neat to see tests!)

backend/coreapp/views.py
nanaian: The idea behind returning the new, logged-out (or logged-in) user is to make the API more RESTful (i.e., you POST to /user to change your current user, and it returns the updated state). I've changed it to return a `message` string with "Login success" or "Logout success" alongside the new `user`.

zbanks: Looks good!

zbanks (approved):

ethteck (approved):

Issue #44, opened 2021-08-05
Show error/warning squiggles on sourcecode + problems view


bates64:
The frontend should attempt to parse warnings/errors emitted by the compiler on specific lines and display warning/error squiggles on those lines. We should also show warnings/errors in their own box similar to vscode, as functions can get lengthy,

See https://microsoft.github.io/monaco-editor/api/modules/monaco.editor.html#setmodelmarkers

nanaian:
Note that different compilers report errors/warnings differently. We only need to grab look for "src.c" and the line number at a base level; we could potentially grab the column number with IDO also.

Gcc281 (only lines 2 and 5 are important)
```
src.c: In function `func':
src.c:2: `non_existent_t' undeclared (first use in this function)
src.c:2: (Each undeclared identifier is reported only once
src.c:2: for each function it appears in.)
src.c:2: parse error before `foo'
```

Ido71 and Ido53:
```
cfe: Error: src.c, line 2: Syntax Error
     non_existent_t foo;
 -------------------^
```

nanaian:
Opinions wanted - should be ignore warnings originating from the context?

ethteck:
I think this is basically accomplished via the language server

PR #122, opened 2021-09-04
Rewrite scratch component

### LINKED ISSUE #110: Scratch page hangs on invalid ID
https://decomp.me/scratch/CBss8 -> invalid, but the page just hangs. We should tell the user this id is invalid.

nanaian:
This PR includes what is effectively a rewrite of the Scratch component, as it was incredibly complex before. It behaves largely the same, except:

- If there are no unsaved changes, the browser will periodically request the scratch data again and update the UI if it has changed (that is, you can watch someone decomp something live!)
- It handles 404 errors correctly (resolves #110)
- It doesn't look as nice when initially loading (I plan to redo the skeleton system in a future PR)
- There isn't a "hide warnings" checkbox anymore; warnings won't be shown at all unless compilation totally fails (#44, when implemented, would have made this checkbox obsolete)
- The author is now shown as "you" or "anon" if an anonymous user owns it, rather than not showing any author at all
- The _Save_ and _Fork_ buttons have been moved to the right of the scratch metadata (author, parent link)
- The _Save_ button is disabled when there are no unsaved changes, rather than adding an asterisk to the button label
- The _Compile_ button no longer spins when compiling
- _Save_, _Compile_, and _Fork_ all display "Loading..." whilst their requests are being sent (they use the new AsyncButton, which I will make prettier later). They also pop up with the error message if something goes wrong during their request / the server doesn't return OK
- Minor backend change to `/compile` to return the compilation under a "compilation" object


ethteck:
I haven't tested yet, but the small backend change looks good. As for the changes you listed, the only thing I think is that it would be nice to be able to see the warnings from the compiler. We may want to eventually add a "logger" panel or something that shows compile errors / warnings if requested. I think it'd be cool to have them show up in the UI, but we shouldn't make it so you can't see a list of them either. For this PR it's fine I think, but just for the future

nanaian:
> We may want to eventually add a "logger" panel or something that shows compile errors / warnings if requested

This should be fairly trivial to add using the same logic as the squiggles - I'll add it to the issue.

ethteck:
Outstanding issues so far:
* When you choose the compiler for a scratch and it's loading the compilation, the diff window says ERROR: compiler not found. We should probably just blank its output until the first compilation result
* The save button seems to never be disabled anymore
* When I fork a scratch and refresh the page, I'm no longer the owner

ethteck:
As for the periodic update function, I like it, but I'd prefer we request the last updated timestamp and only do a full re-request of all the scratch data if the timestamp is newer. I can work on an API call for retrieving the update timestamp of a scratch, if that works for you

nanaian:
I can't reproduce either of these:

> - The save button seems to never be disabled anymore
> - When I fork a scratch and refresh the page, I'm no longer the owner

zbanks:
> > * When I fork a scratch and refresh the page, I'm no longer the owner

I think both of these issues are related to the cookie issues I mentioned on Discord, and are an issue with the dev environment and exist on `main`. 

I did encounter an issue where Ctrl+S on a scratch I do not own raises an `Unhandled Runtime Error`.

nanaian:
@zbanks I've written up the nginx-proxy stuff that circumvents the cookie issue in #131.

> I did encounter an issue where Ctrl+S on a scratch I do not own raises an Unhandled Runtime Error.

I can't reproduce this :(

ethteck (approved):

PR #134, opened 2021-09-21
Fix duplicated line numbers by using new version of `asm_differ`

zbanks:
Pretty small change. Imported fixes from `asm_differ`, then disabled `show_line_numbers` & `show_source` in the config.

Result screenshot: 
2021-09-21-183021_805x225_scrot

### DIFF
--- a/frontend/src/diff/Diff.module.css
+++ b/frontend/src/diff/Diff.module.css
@@ -33,6 +33,7 @@
 .source_filename { font-weight: bold }
 .source_function { font-weight: bold; text-decoration: underline }
 .source_other { font-style: italic }
+.source_line_num { font-style: italic }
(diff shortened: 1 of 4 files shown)


frontend/src/diff/Diff.module.css
```diff
@@ -33,6 +33,7 @@
 .source_filename { font-weight: bold }
 .source_function { font-weight: bold; text-decoration: underline }
 .source_other { font-style: italic }
+.source_line_num { font-style: italic }
```
zbanks, line 36: This technically isn't needed. With `show_line_numbers=False` there shouldn't be any text formatted with the `source_line_number` format, but I'm including it for completeness.

ethteck (approved):

PR #136, opened 2021-09-23
Initial scratch bugfix

### LINKED ISSUE #135: Removing and then restoring source code fails to restore the diff result
https://decomp.me/scratch/UpZ6Q

Remove the first function here and notice the diff change. Then try to add it back and the diff does not update.

ethteck:
Fixes #135 

also makes it a lot less annoying to do local development, changing cookie settings in debug mode


frontend/src/api.ts
```diff
@@ -259,7 +259,9 @@ export function useScratch(slugOrUrl: string): {
     const setScratch = useCallback((partial: Partial<Scratch>) => {
         const scratch = Object.assign({}, localScratch, partial)
 
-        if (!dequal(scratch, savedScratch)) {
+        if (dequal(scratch, savedScratch)) {
```
nanaian, line 262: The reason for this was to not force a rerender if the saved scratch and the serverside scratch are already equal. It's weird to me that this would be the cause of a bug, and this "fix" certainly seems logically wrong.

nanaian: Try testing the live viewing of someone else editing a scratch, this change probably breaks it

PR #140, opened 2021-10-05
Add startup time to hash so the cache is effectively cleared on restart

ethteck:


nanaian (approved):
Good idea, ~~we should consider deleting the cache files also?~~ (Discord discussion)

PR #151, opened 2021-10-08
Diff from label

bates64:


ethteck:
So regarding this PR, I want to make scratches get an arch upon creation. There are a couple of different things to consider:

1. Will arch specification be mandatory? If so, this will break the permuter's integration. If not, do we assume mips if unspecified? That may have downstream issues. I think the best thing to do would be to assume mips currently, and later add an arch detector which tries to pick the "best" option by filtering assembly languages given the instructions the user provided.

2. This also means we probably need a way to change the arch of a scratch after generation. This wouldn't allow the user to change the assembly text, but it would affect the assembly of the provided code. So a scratch would store an Asm object (the text) and not an Assembly (which is the result of assembling an Asm)

PR #187, opened 2021-10-19
Add GameCube & Wii compiler support

### LINKED ISSUE #74: Add mwcc (powerpc) compilers
These won't be in the repo, but they will become an option if their binaries are found

mkst:
Closes #74

Latest:

image

If anyone wants to do something different/better with the logo, I'm all ears. I'm not a graphics person :disappointed: 

New question:
- What to do about *patched* MWCC compilers, e.g. Twilight Princess which has 1-byte patch compared to 'vanilla' `2.4.7 build 108 (GC MW 2.7)`. Do we want both?

TODO:
- ~~fixup tokenising of checkbox values~~
- how to support:
```js
        <Checkbox flag="-pragma 'check_header_flags off'" description="-pragma 'check_header_flags off'" />
        <Checkbox flag="-pragma 'force_active on'" description="-pragma 'force_active on'" />
```
i.e. arguments with quotes.

--------------
OLD

Open question: what should the friendly name be for these compilers? `MWCC 1.0 (Version 2.3.3 build 144)` or `GC MW 1.0 (Version 2.3.3 build 144)` or something else?

Expects the `powerpc-eabi-*` binaries to be in your path (i.e. get devkitpro, install to /opt/devkitpro/ and then `cp /opt/devkitpro/devkitPPC/bin/* /usr/bin/`)...

If you want me to add any more compilers, just give me the version numbers and I'll add them.

image

### DIFF
--- a/frontend/src/components/compiler/compiler_settings/ido.tsx
+++ b/frontend/src/components/compiler/compiler_settings/ido.tsx
@@ -1,0 +1,27 @@
+import { Checkbox, FlagSet, FlagOption } from "../CompilerOpts"
+
+export function CommonIDOFlags() {
+    return <>
+        <FlagSet name="Optimization level">
+            <FlagOption flag="-O0" description="No optimization" />
+            <FlagOption flag="-O1" description="Some optimization" />
+            <FlagOption flag="-O2" description="Standard optimization" />
+            <FlagOption flag="-O3" description="Heavy optimization" />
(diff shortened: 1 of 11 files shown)


TakaRikka:
Twilight Princess decomp is currently using a modified version of 2.7 if you want to add that. the binary is pinned in the tp-decomp channel in zeldaret server

PR #193, opened 2021-10-29
Scratch fixes

nanaian:


ethteck:
scrixes

frontend/src/components/compiler/CompilerOpts.tsx
```diff
@@ -55,58 +58,41 @@ export function FlagOption({ flag, description }: { flag: string, description?:
...
-            compiler,
-            compiler_flags: opts,
-        })
-    }
-
+export default function CompilerOpts({ compiler, flags, onCompilerChange, onFlagsChange, platform, title, isPopup }: Props) {
     return <OptsContext.Provider value={{
         checkFlag(flag: string) {
-            return (" " + opts + " ").includes(" " + flag + " ")
+            return flags.split(" ").includes(flag)
```
mkst, line 74: Doesn't this break with args that are multi part like `- foo 123`? This is why I did the whitespace approach

nanaian: Ah, yes, good point

ethteck:
Known issues (as mentioned in discord):

- on main, I can save the scratches I create immediately, whereas on scrixes I cannot 
- on main, saving seems to convince the front-end that the saved state is equal to the current state, so the save button isn't pressable again until other updates. on scrixes, I can keep pressing save
- on main, when I fork, I don't get the "unsaved changes" dialog that I get on scrixes, and when I move to the new fork, changes to the source code trigger a state change. on scrixes, the state seems kind of frozen after forking

PR #195, opened 2021-10-31
Add /export endpoint for downloading .zip

zbanks:
Not referenced on the frontend at all, mostly exists as a proof-of-concept for building a `.zip` file dynamically from the files in the database.

### DIFF
--- a/backend/coreapp/urls.py
+++ b/backend/coreapp/urls.py
@@ -9,7 +9,7 @@ urlpatterns = [
     path('scratch/<slug:slug>/compile', views.compile, name='scratch-compile'),
     path('scratch/<slug:slug>/claim', views.ScratchClaim.as_view(), name='scratch-claim'),
     path('scratch/<slug:slug>/fork', views.fork, name='scratch-fork'),
-    path('scratch/<slug:slug>/export', views.export, name='scratch-export'),
+    path('scratch/<slug:slug>/export', views.ScratchExport.as_view(), name='scratch-export'),
     path('user', views.CurrentUser.as_view(), name="current-user"),
     path('users/<slug:username>', views.user, name="user-detail"),
 ]
--- a/backend/coreapp/urls.py
+++ b/backend/coreapp/urls.py
@@ -9,6 +9,7 @@ urlpatterns = [
     path('scratch/<slug:slug>/compile', views.compile, name='scratch-compile'),
     path('scratch/<slug:slug>/claim', views.ScratchClaim.as_view(), name='scratch-claim'),
     path('scratch/<slug:slug>/fork', views.fork, name='scratch-fork'),
+    path('scratch/<slug:slug>/export', views.export, name='scratch-export'),
     path('user', views.CurrentUser.as_view(), name="current-user"),
     path('users/<slug:username>', views.user, name="user-detail"),
 ]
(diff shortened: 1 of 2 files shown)


backend/coreapp/views.py
```diff
@@ -337,3 +343,29 @@ def user(request, username):
...
     return Response(serialize_profile(request, get_object_or_404(Profile, user__username=username)))
+
+def export(request, slug):
+    # Double-check `slug` to prevent some injection attacks
+    validate_slug(slug)
+    scratch = get_object_or_404(Scratch, slug=slug)
+
+    metadata = ScratchWithMetadataSerializer(scratch, context={ "request": request }).data
+    metadata.pop("source_code")
+    metadata.pop("context")
```
zbanks, line 354: Not sure if I should create a new `Serializer` for this specific case?

nanaian: I think this is fine

backend/coreapp/views.py
```diff
@@ -337,3 +343,29 @@ def user(request, username):
     """
 
     return Response(serialize_profile(request, get_object_or_404(Profile, user__username=username)))
+
+def export(request, slug):
```
nanaian, line 347: Can you throw the last_modified/etag stuff that scratch api uses on this endpoint?

nanaian: This endpoint should also be limited to get requests

backend/coreapp/urls.py
```diff
@@ -9,6 +9,7 @@
     path('scratch/<slug:slug>/compile', views.compile, name='scratch-compile'),
     path('scratch/<slug:slug>/claim', views.ScratchClaim.as_view(), name='scratch-claim'),
     path('scratch/<slug:slug>/fork', views.fork, name='scratch-fork'),
+    path('scratch/<slug:slug>/export', views.export, name='scratch-export'),
```
nanaian, line 12: I wonder if it would be better to expose exporting as
```
GET /api/scratch/slug
Accept: application/zip
```
(i.e. export when the requested content-type isnt json)

nanaian (changes requested):

backend/coreapp/urls.py
zbanks: I think that might be difficult to use in practice? 

The way it is - it can be opened with a plain link. I don't think there's a way from JS to tell the browser to open a link with a certain `Accept` header? I think you'd have to XHR the entire data, and then generate a "fake" download link?

backend/coreapp/views.py
zbanks: Refactored, and moved `scratch_condition` outside of the `ScratchDetail` class to share it?

backend/coreapp/urls.py
ethteck: yeah I'd rather opt for user-friendliness over a more "proper" design

backend/coreapp/views.py
nanaian: ty

ethteck (approved):

backend/coreapp/urls.py
nanaian: @ethteck This is the API, it's not intended to be used by users (and hence correctness is more important to me).

@zbanks good point, nvm. I'm a bit wary of this being under /api though, we can route /scratch/asdf.zip to the API with nextjs I think.

nanaian (approved):

PR #199, opened 2021-11-10
Possibly fix refresh bug? (suspense issue?)

zbanks:
I don't fully understand the issue here - as Ethan was saying in the Discord, `localScratch` always seems `undefined` inside `updateLocalScratch` -- even *after* it has been set, even several renders later. (e.g. it's not `undefined` outside of that fn).


nanaian (approved):
Oh gosh, this looks entirely plausible: `useCallback` doesn't perform a deep comparison! I forgot the setter functions could take a function taking the previous.

PR #200, opened 2021-11-13
Fix 2 bugs

ethteck:


frontend/src/components/compiler/PresetSelect.tsx
```diff
@@ -66,13 +66,12 @@ export const PRESETS = [
...
 
-export default function PresetSelect({ className, platform, compiler, opts, setCompiler, setOpts, serverCompilers }: {
+export default function PresetSelect({ className, platform, compiler, opts, setPreset, serverCompilers }: {
     className?: string
     platform: string
     compiler: string
     opts: string
-    setCompiler: (compiler: string) => void
-    setOpts: (opts: string) => void
+    setPreset: (preset: { name: string, compiler: string, opts: string }) => void
```
nanaian, line 74: Can you add a `type CompilerPreset = { name: string, compiler: string, opts: string }` and use this here and in the other places you had this anonymous type

nanaian (changes requested):

nanaian (approved):

PR #210, opened 2021-11-18
Hide wine's warnings

mkst:
So, the changes:

1. Add a `/wine` directory to the docker image, run `winecfg` to populate it with wine related stuff
2. Default WINEPREFIX dir to `/tmp/wine` and create it if it does not exist (otherwise sandbox mount fails)
3. Move the `WINEPREFIX=/wine` env from compiler_wrapper into the sandbox code
4. bindmount `/etc/fonts` because wine complains about not being able to find font config
5. expose `WINEPREFIX` in settings

BTW if you get the "wine: too many open files", edit `/etc/security/limits.conf` and add these two lines:
```
*               soft    nofile          65535
*               hard    nofile          65535
```

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -62,16 +62,9 @@ RUN if [ "${ENABLE_PS1_SUPPORT}" = "YES" ] ; then \
     fi
 
 # gc/wii specifics
+COPY --from=devkitpro/devkitppc:20210726 /opt/devkitpro/devkitPPC/bin/powerpc* /usr/bin/
 RUN if [ "${ENABLE_GC_WII_SUPPORT}" = "YES" ] ; then \
-    bash /compilers/mwcc.sh && \
-    wget https://github.com/devkitPro/pacman/releases/download/v1.0.2/devkitpro-pacman.amd64.deb \
-        -O devkitpro.deb && \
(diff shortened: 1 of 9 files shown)


backend/Dockerfile
```diff
@@ -51,7 +51,9 @@ RUN if [ "${ENABLE_PS1_SUPPORT}" = "YES" ] || [ "${ENABLE_GC_WII_SUPPORT}" = "YE
     dpkg --add-architecture i386 && apt-get update && \
     apt-get install -y -o APT::Immediate-Configure=false \
         wine; \
-    fi
+    fi && \
+    mkdir -p /wine && \
+    WINEPREFIX=/wine winecfg
```
zbanks, line 56: I think `mkdir -p /wine` should always be run, but `WINEPREFIX=/wine winecfg` should only be run in the body of the `if`? (Maybe move the `mkdir` before the `if`?)

mkst: good point!

zbanks (approved):

ethteck:
getting `wine: error opening file descriptor for wineserver tmpdir file` dockerless, locally. it'd be good to add a test that invokes wine

PR #211, opened 2021-11-19
2 more things

### LINKED ISSUE #188: Clarify purpose of "function name" in scratch creation UI and fix potential issues
It's not clear that the auto-populated "function name" is actually going to be the starting point of the diff, as seen by Maide when working on MM. I think we should call this "diff origin / function name" or something like that. Also, the auto-population regex should probably ignore the common branch label pattern: `L[0-9A-Fa-f]{8}`

### LINKED ISSUE #207: Address disk space issue
The elf_object in `Compilation` takes up a ton of space. We don't need to keep all of them around all the time. I think on bootup of the server, we should clear any elf_object entries that are older than x days, as a starting point

ethteck:
Ram cache for elf_object (fixes #207) and better diff label detection / explanation (fixes #188)


ethteck:
No clue how to fix the mypy issue - advice wanted

nanaian:
> No clue how to fix the mypy issue - advice wanted

Probably mypy-ignore that line, mypy is bad at decorators it seems

nanaian (approved):

backend/coreapp/compiler_wrapper.py
```diff
@@ -239,15 +245,18 @@ def get_objdump_command(platform: str) -> Optional[str]:
...
 def _check_assembly_cache(*args: str) -> Tuple[Optional[Assembly], str]:
     hash = util.gen_hash(args)
     return Assembly.objects.filter(hash=hash).first(), hash
 
 
+class CompilationResult:
+    def __init__(self, compilation: Optional[Compilation], elf_object: bytes, errors: str):
+        self.compilation = compilation
+        self.elf_object = elf_object
+        self.errors = errors
```
zbanks, line 257: This would be a great use of `@dataclass`, like `Platform` in `compiler_wrapper.py`

backend/coreapp/compiler_wrapper.py
```diff
@@ -329,20 +338,15 @@ def filter_compiler_flags(compiler: str, compiler_flags: str) -> str:
         return " ".join(flags)
 
     @staticmethod
-    def compile_code(compiler: str, compiler_flags: str, code: str, context: str, to_regenerate: Optional[Compilation] = None) -> Tuple[Optional[Compilation], Optional[str]]:
+    @lru_cache(maxsize=settings.COMPILATION_CACHE_SIZE, typed=True)
```
zbanks, line 341: You probably don't need `typed=True`? All the args are always `str`?

From [`lru_cache` docs](https://docs.python.org/3/library/functools.html#functools.lru_cache)
> If typed is set to true, function arguments of different types will be cached separately. If typed is false, the implementation will usually regard them as equivalent calls and only cache a single result. (Some types such as str and int may be cached separately even when typed is false.)

zbanks (changes requested):

backend/coreapp/compiler_wrapper.py
ethteck: Ah, I figured it was just extra cautious, but yeah in practice it won't matter

PR #216, opened 2021-11-22
Add calculation and display of match %

ethteck:


frontend/src/components/ScoreBadge.tsx
```diff
@@ -3,11 +3,24 @@ import classNames from "classnames"
 
 import styles from "./ScoreBadge.module.scss"
 
+function round(num: number) {
+    return Math.round((num + Number.EPSILON) * 100) / 100
```
nanaian, line 7: ELI5 pls? In a comment, perhaps?

nanaian (changes requested):
Use camelCase for max_score (maxScore) for all but the `DiffOutput` type coming from the API

frontend/src/components/ScoreBadge.tsx
mkst: `// force gud rounding https://stackoverflow.com/a/11832950/7290888`

nanaian: update: use toFIxed(2) to convert to string instead of rounding then converting that

frontend/src/components/ScoreBadge.tsx
```diff
@@ -3,11 +3,20 @@ import classNames from "classnames"
 
 import styles from "./ScoreBadge.module.scss"
 
+function calculateScorePercent(score: number, maxScore: number) {
+    if (score > maxScore) {
+        return 0
```
nanaian, line 8:
```suggestion
function calculateScorePercent(score: number, maxScore: number): string {
    if (score > maxScore) {
        return "0"
```

nanaian (changes requested):
types

frontend/src/components/ScoreBadge.tsx
ethteck: I disagree that this should be returning a string

ethteck: overruled

nanaian (approved):

PR #218, opened 2021-11-24
update Pikmin and Pikmin 2 presets

EpochFlame:
also remove unnecessary compiler logic/config for an obsolete compiler fork


EpochFlame:
should be ready for testing

PR #228, opened 2021-12-06
Update 1.2.5e to use frank.py

mkst:
image

Until 1.2.5 can be patched to match expected behaviour, this is the best solution at the time of writing...


EpochFlame:
all hail frank

Issue #238, opened 2021-12-20
Add a dummy compiler


nanaian:
Add a dummy arch and compiler for simple-dev-environment purposes (e.g. always emits `jr $ra`).

Related to #153

nanaian:
A dummy compiler should also be used in most tests

PR #246, opened 2021-12-22
Dummy compiler + better error handling and reporting

### LINKED ISSUE #160: Return consistent errors from backend
- All error responses should come in the form of APIException subclasses
- Override the rest framework APIException handler to return `{ "error": "some_code", "message": "Detailed message" }` instead of `{ "detail": "Detailed message" }`
- Failed compiles should return error 400

ethteck:
fixes #160, #238


backend/coreapp/asm_diff_wrapper.py
```diff
@@ -120,38 +124,34 @@ def diff(target_assembly: Assembly, platform: str, diff_label:Optional[str], com
...
         try:
             # TODO: It would be nice to get a python object from `run_diff()` to avoid the
             # JSON roundtrip. See https://github.com/simonlindholm/asm-differ/issues/56
             result = json.loads(display.run_diff()[0])
             result["error"] = None
-        except Exception:
-            logger.exception("Error running asm-differ")
-            return {"error": "Error running asm-differ"}
+        except Exception as e:
+            raise DiffError(f"Error running asm-differ: {str(e)}")
```
mkst, line 155: What happens without the str() for these exceptions? Is string representation the default?

backend/coreapp/compiler_wrapper.py
```diff
@@ -339,8 +348,7 @@ def filter_compiler_flags(compiler: str, compiler_flags: str) -> str:
     @lru_cache(maxsize=settings.COMPILATION_CACHE_SIZE) # type: ignore
     def compile_code(compiler: str, compiler_flags: str, code: str, context: str) -> CompilationResult:
         if compiler not in _compilers:
-            logger.debug(f"Compiler {compiler} not found")
-            return CompilationResult(b'', "ERROR: Compiler not found")
+            raise CompilationError("Compiler not found")
```
mkst, line 351: "Compiler {compiler} not found" ?

backend/coreapp/asm_diff_wrapper.py
ethteck: I'm not sure...this is what was suggested to me by copilot. @zbanks do you know if this is preferred / safe, or it's okay to just use e directly?

backend/coreapp/tests.py
```diff
@@ -229,7 +225,12 @@ def test_mwcc_wine(self):
         """
         result = CompilerWrapper.compile_code("mwcc_247_92", "-str reuse -inline on -fp off -O0", "int func(void) { return 5; }", "extern char libvar1;\r\nextern char libvar2;\r\n")
 
-        self.assertGreater(len(result.elf_object), 0, "The compilation result should be non-null")
+    def test_dummy_compiler(self):
+        """
+        Ensure basic functionality works for the dummy compiler
+        """
+
+        result = CompilerWrapper.compile_code("dummy", "", "sample text 123", "")
```
mkst, line 234: is no assertion required?

ethteck: If the compilation returns a non-200, the test automatically fails per the test framework

ethteck: or, sorry. this isn't an api call. if the function raises an error (which it will if there are errors), the test will fail *

backend/coreapp/views/scratch.py
```diff
@@ -293,15 +272,25 @@ def compile(request, slug):
     scratch: Scratch = Scratch.objects.get(slug=slug)
 
     # Get the context from the backend if it's not provided
-    if not context:
+    if context == None:
```
mkst, line 275: is None

backend/coreapp/views/scratch.py
```diff
@@ -293,15 +272,25 @@ def compile(request, slug):
...
+    try:
+        result = CompilerWrapper.compile_code(compiler, compiler_flags, code, context)
+    except CompilationError as e:
+        return APIException(str(e), code=str(status.HTTP_400_BAD_REQUEST))
 
-    diff_output: Optional[Dict[str, Any]] = None
-    if result.elf_object:
+    try:
         diff_output = AsmDifferWrapper.diff(scratch.target_assembly, scratch.platform, scratch.diff_label, result.elf_object)
+    except AssemblyError as e:
```
mkst, line 286: unless you want to do something different for them, I think you can do 
```py
except (AssemblyError, DiffError, ObjdumpError, CompilationError) as e:
```

ethteck: yeah I wasn't sure yet if we should handle these differently. For now I'll do that, though

backend/coreapp/asm_diff_wrapper.py
```diff
@@ -93,12 +100,9 @@ def run_objdump(target_data: bytes, platform: str, config: asm_differ.Config, la
                         },
                     )
                 except subprocess.CalledProcessError as e:
-                    logger.error(e)
-                    logger.error(e.stderr)
-                    return None
+                    raise ObjdumpError(f"Error running objdump: {str(e)}")
```
mkst, line 103: Does str(e) include the e.stdout + e.stderr? Wondering if we'd lose any messages that would help identify any errors?

ethteck: I'm not sure...good point

backend/coreapp/asm_diff_wrapper.py
zbanks: It defaults to using the [`__format__` method](https://docs.python.org/3/library/functions.html#format) on the object, instead of `__str__`, though I believe the default implementation of that method should be equivalent to `str(...)` for these exceptions. 
I think `{e}` would be preferred here?

ethteck: great, thanks!

backend/coreapp/views/scratch.py
```diff
@@ -293,15 +272,19 @@ def compile(request, slug):
...
+        result = CompilerWrapper.compile_code(compiler, compiler_flags, code, context)
+    except CompilationError as e:
+        return APIException(str(e), code=str(status.HTTP_400_BAD_REQUEST))
 
-    diff_output: Optional[Dict[str, Any]] = None
-    if result.elf_object:
+    try:
         diff_output = AsmDifferWrapper.diff(scratch.target_assembly, scratch.platform, scratch.diff_label, result.elf_object)
+    except (AssemblyError, DiffError, ObjdumpError, CompilationError) as e:
+        return APIException(str(e))
```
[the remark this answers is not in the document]
ethteck: yeah, thanks

backend/coreapp/asm_diff_wrapper.py
```diff
@@ -13,6 +13,14 @@
...
 MAX_FUNC_SIZE_LINES = 5000
 
+class DiffError(Exception):
+    def __init__(self, message: str):
+        super().__init__(f"Diff error: {message}")
+
+class ObjdumpError(Exception):
+    def __init__(self, message: str):
+        super().__init__(f"Objdump error: {message}")
+
```
zbanks, line 23: It might be helpful to define a base exception class here? Something like this?

```python
class SubprocessError(Exception):
    SUBPROCESS_NAME: ClassVar[str] = "Subprocess"
    stdout: str
    stderr: str

    def __init__(self, message: str):
        super().__init__(f"{self.SUBPROCESS_NAME} error: {message}")
        self.stdout = ""
        self.stderr = ""

    @staticmethod
    def from_process_error(self, ex: CalledProcessError) -> SubprocessError:
        error = SubprocessError(f"{ex.cmd[0]} returned {ex.returncode}")
        error.stdout = ex.stdout
        error.stderr = ex.stderr

class DiffError(SubprocessError):
    SUBPROCESS_NAME: ClassVar[str] = "Diff"

class ObjdumpError(SubprocessError):
    SUBPROCESS_NAME: ClassVar[str] = "Objdump"

# ...

try:
    pass
except subprocess.CalledProcessError as e:
    raise DiffError.from_process_error(e)
```

(You could tweak things to always include stderr/stdout in the message, etc.)

.env
```diff
@@ -1,4 +1,5 @@
 DEBUG=on
+PROD=off
```
nanaian, line 2: Perhaps this should instead be called `DUMMY_COMPILER`?

backend/decompme/settings.py
```diff
@@ -148,6 +157,13 @@
         'handlers': ['console'],
         'level': 'DEBUG',
     },
+    'loggers': {
+        'django': {
+            'handlers': ['console'],
+            'level': os.getenv('DJANGO_LOG_LEVEL', 'INFO'),
```
nanaian, line 163: Should this be `env()` instead

.env
ethteck: I thought there may be other things we want to put behind this flag, like attempting to use other features that shouldn't work locally (permuter@home)

nanaian: I'd rather use separate flags for that. (also why would permuter@home not work in dev? how would we test it?)

backend/decompme/settings.py
```diff
@@ -165,7 +183,7 @@
 LOCAL_FILE_DIR = BASE_DIR / "local_files"
 
 USE_SANDBOX_JAIL = env("USE_SANDBOX_JAIL")
-if sys.platform == "darwin":
+if sys.platform == "darwin" or (sys.platform == "win32" and not in_wsl()):
```
mkst, line 186: Is the `not in_wsl` check redundant here? Is the following more correct?

```py
# always disable sandbox on mac and windows
if sys.platform == "darwin" or sys.platform == "win32"):
```

ethteck: I wasn't sure if wsl still technically reports as being windows in some system properties. If this suffices, then yeah, that's better. I'll change it

PR #247, opened 2021-12-22
MVP Training Implementation

FluentCoding:
Not completely done yet, there are some points that we have to discuss:

- [ ] Should we use the user database to store training progress?
- [ ] Should we host the list of training scratches in the backend or should we keep it in the frontend?


nanaian:
>  Should we use the user database to store training progress?

I think we should add a new model with a foreign key user and the progress of each scenario

> Should we host the list of training scratches in the backend or should we keep it in the frontend?

I think frontend is OK but the fact that scenarios reference specific scratches implies it should have a backend form for admins to add new scenarios. I'm unsure if instead we should have the scenario source_code etc in the repo

FluentCoding:
> I think frontend is OK but the fact that scenarios reference specific scratches implies it should have a backend form for admins to add new scenarios. I'm unsure if instead we should have the scenario source_code etc in the repo

Hm, an admin panel sounds interesting for that imo. We could also do something really hacky and make a scratch that basically acts as the database for the scenarios (e.g. in XML format in a multiline comment:)

```c
/*
<name>Test</name>
<description>Test</description>
<difficulty>easy</difficulty>
<slug>ABCDE</slug>
*/
```

And parse that scratch every time the user needs access to the scenarios.

nanaian:
>  We could also do something really hacky

No thanks :shipit: 

---

I think we need

- Database models for
  - Training scenario
  - User x scenario progress
- Admin dashboard for adding training scenarios

I also pointed out to @ethteck that when we drop project support we can allow project admins to create training scenarios for their project in particular. e.g. papermario can teach how to match script API functions

FluentCoding:
Yeah, I absolutely agree, the scratch as a database suggestion was more of a joke tbh lol. Project support sounds sick btw, I would look forward into adding support for adding training scenarious to projects.

Also, how would we implement admin dashboard authentication? Add a list of emails to an admin role that has access to decomp.me/admin e.g.?

ethteck:
My thoughts on the DBs:

### Course (has a 1-many relationship with CourseStep)
name - the name of the course
description - the description of the course
project - stub for now

### CourseStep
name - the name of the step (could use scratch's name but feels better to store its own name in its object. plus this allows not worrying about the scratch changing and suddenly the name is different, etc)
stepNumber - which step in the course is this (will be used to gauge completion)
course - a reference to the Course it is part of
scratch - a Scratch which is the starting point for the step in terms of code / asm / context
content - some html or script or other kind of data which is the "course material" (probably best to start as just text for now)

### CourseProgress - a row here represents the progress a user has made on a CourseStep
user - the User
courseStep - the CourseStep
scratch - the (latest?) fork of the CourseStep's starting point that represents the user's workspace for this step
complete - a bool representing the completion of the course - this can also be a percentage if we want to go that route. this also allows the user to restart the course but maintain their completion progress in it

To determine user john's progress on the n65 course, we can do a query that looks like `CourseProjects.objects.filter(user=john).filter(courseStep.course=n65)` - we then can see the completion of each step of the course john has made and return these to the front-end to show completion as well as scratch links to go and continue working on something or revisit what you did earlier.

frontend/src/components/Scratch/RenderTabs.tsx
```diff
@@ -0,0 +1,135 @@
...
+ * Renders the left tabs of the scratch
+ * @param {Array<LeftScratchTab>} [filter=undefined] The tabs you want to render
+ * @returns Left tabs of scratch
+ */
+export function renderLeftTabs({ scratch, setScratch }: {
+    scratch: Scratch
+    setScratch: (s: Partial<Scratch>) => void
+}, filter?: Array<LeftScratchTab>): React.ReactElement<typeof Tab>[] {
+    const sourceEditor = useRef<EditorInstance>() // eslint-disable-line react-hooks/rules-of-hooks
+    const contextEditor = useRef<EditorInstance>() // eslint-disable-line react-hooks/rules-of-hooks
```
nanaian, line 44: We might be able to avoid these lint disables by naming this function `useLeftTabs` or something.

frontend/src/components/Scratch/RenderTabs.tsx
```diff
@@ -0,0 +1,135 @@
...
+    SETTINGS,
+}
+
+export enum RightScratchTab {
+    DIFF,
+}
+
+/**
+ * Renders the left tabs of the scratch
+ * @param {Array<LeftScratchTab>} [filter=undefined] The tabs you want to render
```
nanaian, line 36: Can you explain this doc comment to me?

frontend/src/components/Scratch/RenderTabs.tsx
```diff
@@ -0,0 +1,135 @@
...
+        return Object.values(tabs)
+    else
+        return Object.keys(tabs).filter(key => filter.includes(+key)).map(key => tabs[key])
+}
+
+export enum LeftScratchTab {
+    SOURCE_CODE,
+    ABOUT,
+    CONTEXT,
+    SETTINGS,
```
nanaian, line 27: Good call.

frontend/src/components/Scratch/ScratchBody.tsx
```diff
@@ -0,0 +1,56 @@
...
+export default function ScratchBody({
+    container,
+    leftTab,
+    rightTab,
+    setLeftTab,
+    setRightTab,
+    leftTabs,
+    rightTabs,
+    setCompilerOpts,
+    scratch,
```
nanaian, line 20: We should type these props.

FluentCoding: Yep, forgot it!

frontend/src/components/Scratch/RenderTabs.tsx
FluentCoding: Maybe "render" was the wrong wording for this, but basically "Returns the left tabs of the scratch layout".
The second line also says that filter can be undefined (filter=undefined), i think thats the way you do it in JSDoc.

nanaian (approved):

ethteck (approved):

ethteck (changes requested):
could you silence the warnings so the build passes?

PR #256, opened 2021-12-28
Improve diff layout

nanaian:
Builds on #252:

- Fixes layout on mobile; thin viewports will not cause the columns to compress (notice horizontal scrollbar in image)

image

- Reverts to having the columns being left-justified, instead of 50-50

image

- Various style changes e.g. adds lines between the columns


ethteck:
L👀ks g👀d

Issue #75, opened 2021-08-19
Add clang (Switch) compiler support


ethteck:
Questions:
* Are there multiple versions we should care about?
* What are the common flags?
* How is it invoked?
* What dependencies do we need to worry about?
* Do we build it as part of project setup, include / obtain binaries, or leave it up to the instance runner to provide the compiler?

leoetlino:
> * Are there multiple versions we should care about?

Yes, different games can use different versions of LLVM. Even a single game can use several compiler versions because Nintendo's toolchain team is pretty good at staying up-to-date with upstream LLVM.

> * What are the common flags?

The target flags: `-target aarch64 -mcpu=cortex-a57+fp+simd+crypto+crc -fPIC -stdlib=libc++`

Games sometimes pass flags that affect codegen though, e.g. `-g` (debug info), `-mno-implicit-float`, `-O2` vs `-O3`, etc. Given the number of flags I'm not sure only offering presets is a good idea because you'd have 10 (versions) × 2 (optimization flags) × 2 (debug info flags) × 2 (float codegen flags) × ... possibilities.

> * How is it invoked?

Well it's Clang, so it has the same standard interface as GCC.

> * What dependencies do we need to worry about?

None (other than system libs). Official Clang binary builds statically link LLVM.

> * Do we build it as part of project setup, include / obtain binaries, or leave it up to the instance runner to provide the compiler?

While possible (unlike the vast majority of other decomp projects!), building Clang from source would take a significant amount of time because LLVM has a huuuuuuuuuuuuuge codebase. If you need Clang binaries I'd recommend just grabbing them from https://releases.llvm.org/ -- the LLVM project provides official pre-built releases for every version from 1.0 to the latest 12.0 (again a luxury that most decomp projects do not have!)

PR #265, opened 2021-12-30
Copy scratch name to fork

### LINKED ISSUE #175: Use "Untitled fork of (scratch)" for fork name placeholder
Instead of "Untitled scratch".

e.g. a fork of a scratch called "func_123" can placehold to "Untitled fork of func_123"
a fork of an unnamed scratch can just be "Untitled fork"

ethteck:
I think this is an ideal way of tackling #175) as we already know a fork is a fork via the ui. adding "fork of" is just going to get spammy I think

closes #175

### DIFF
--- a/backend/coreapp/serializers.py
+++ b/backend/coreapp/serializers.py
@@ -39,6 +39,7 @@ class ProfileField(ProfileFieldBaseClass):
         return serialize_profile(self.context["request"], profile)
 
 class ScratchCreateSerializer(serializers.Serializer[None]):
+    name = serializers.CharField(allow_blank=True, required=False)
     compiler = serializers.CharField(allow_blank=True, required=True)
     platform = serializers.CharField(allow_blank=True, required=False)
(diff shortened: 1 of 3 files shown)


nanaian (changes requested):
Frontend needs to not provide `name` to the fork endpoint for this to work, so please add that (line 277, api.ts). I think this can be done by doing `Object.assign({}, ..., { name: undefined })`

nanaian:
Also the relevant issue does not appear to be #75 but #175

ethteck:
> Frontend needs to not provide `name` to the fork endpoint for this to work, so please add that (line 277, api.ts). I think this can be done by doing `Object.assign({}, ..., { name: undefined })`

how do I gracefully renove name from the result of Object.assign?

nanaian:
`Object.assign({}, scratch, whatever, { name: undefined })` - assign updates with precedence to the right. Or use the `delete` operator.

Thinking about this further it might Just Work without any frontend changes? Please test on the frontend for me :pray:

PR #268, opened 2021-12-31
Add platform to scratch

nanaian:
- remove gaps from n64 platform logo
- split PlatformIcon into its own component
- platform icon in scratch toolbar
- add platform name to AboutScratch


ethteck:
Looks good, but the ps/dummy icons are giant when you make a scratch for those

PR #270, opened 2021-12-31
Add settings page

### LINKED ISSUE #61: "automatically compile" checkbox that is on by default but can be unchecked to disable automatic recompilation

nanaian:
- nav user menu popup
- add settings page (fixes #61)

Note that settings are stored locally for now, rather than on the server per-user. I think this is fine for now


frontend/src/pages/u/[username].tsx
```diff
@@ -1,12 +1,11 @@
 import { GetStaticProps } from "next"
 
 import Image from "next/image"
-import { useRouter } from "next/router"
+//import { useRouter } from "next/router"
```
Comment (mkst) [inline frontend/src/pages/u/[username].tsx:4]: remove?

Comment (nanaian) [reply frontend/src/pages/u/[username].tsx:4]: ty

Issue #272, opened 2021-12-31
Give all serialised objects a "url" field that points to the API endpoint for themselves


nanaian:
Work towards this is in the `scratch-viewset` branch

PR #274, opened 2022-01-01
Prevent profiles being created for SSR requests, serializer cleanup

### LINKED ISSUE #184: Next.js SSR creates a new profile for every request
Next.js' server-side rendering doesn't hold cookies between requests, so we end up with
image

Anonymous profiles are never deleted (an issue in of itself). I imagine Next.js sends over a header that identifies it as Next.js, we should detect that in django and delete the anonymously made profile after the request is handled, or not make a profile at all.

ethteck:
Fixes #184

### DIFF
--- a/backend/coreapp/middleware.py
+++ b/backend/coreapp/middleware.py
@@ -34,7 +34,7 @@ def set_user_profile(get_response):
     """
 
     def middleware(request: Request):
-        # Skip if the reuqest is from SSR
+        # Skip if the request is from SSR
         if "User-Agent" in request.headers and "node-fetch" in request.headers["User-Agent"]:
             request.profile = Profile()
             return get_response(request)
--- a/backend/coreapp/middleware.py
+++ b/backend/coreapp/middleware.py
@@ -34,6 +34,11 @@ def set_user_profile(get_response):
     """
 
     def middleware(request: Request):
+        # Skip if the reuqest is from SSR
+        if "User-Agent" in request.headers and "node-fetch" in request.headers["User-Agent"]:
+            request.profile = Profile()
+            return get_response(request)
+
         profile: Optional[Profile] = None
 
         # Use the user's profile if they're logged in
--- a/backend/coreapp/serializers.py
+++ b/backend/coreapp/serializers.py
@@ -51,8 +51,13 @@ class ScratchCreateSerializer(serializers.Serializer[None]):
 
 class ScratchSerializer(serializers.ModelSerializer[Scratch]):
     owner = ProfileField(read_only=True)
+    parent = serializers.HyperlinkedRelatedField( # type: ignore
+        read_only=True,
+        view_name="scratch-detail",
+        lookup_field="slug",
+    )
 
     class Meta:
         model = Scratch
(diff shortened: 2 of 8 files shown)


backend/coreapp/middleware.py
```diff
@@ -34,6 +34,11 @@ def set_user_profile(get_response):
     """
 
     def middleware(request: Request):
+        # Skip if the reuqest is from SSR
```
mkst, line 37: request

backend/coreapp/middleware.py
```diff
@@ -34,6 +34,11 @@ def set_user_profile(get_response):
     """
 
     def middleware(request: Request):
+        # Skip if the request is from SSR
+        if "User-Agent" in request.headers and "node-fetch" in request.headers["User-Agent"]:
```
nanaian, line 38: I can see this potentially causing issues in the future wrt node-fetch-based API consumers, but we don't have any so they can complain later.

backend/coreapp/serializers.py
```diff
@@ -50,19 +50,9 @@ class ScratchCreateSerializer(serializers.Serializer[None]):
...
 
 class ScratchSerializer(serializers.ModelSerializer[Scratch]):
-    class Meta:
-        model = Scratch
-        fields = ["slug", "name", "description", "compiler", "platform", "compiler_flags", "target_assembly", "source_code", "context", "diff_label", "score", "max_score"]
-
-# XXX: ideally we would just use ScratchSerializer, but adding owner and parent breaks creation
-class ScratchWithMetadataSerializer(serializers.ModelSerializer[Scratch]):
     owner = ProfileField(read_only=True)
-    parent = serializers.HyperlinkedRelatedField( # type: ignore
```
nanaian, line 60: What is the shape of `parent` now this is removed?

backend/coreapp/serializers.py
```diff
@@ -50,19 +50,9 @@ class ScratchCreateSerializer(serializers.Serializer[None]):
...
-        read_only=True,
-        view_name="scratch-detail",
-        lookup_field="slug",
-    )
 
     class Meta:
         model = Scratch
-        fields = ["slug", "name", "description", "compiler", "platform", "compiler_flags", "source_code", "context", "owner", "parent", "diff_label", "score", "max_score"]
+        fields = ["slug", "name", "description", "compiler", "platform", "compiler_flags", "target_assembly", "source_code", "context", "diff_label", "score", "max_score", "parent", "owner"]
+        read_only_fields = ["slug", "parent"]
```
nanaian, line 58: I think owner should go here jic

backend/coreapp/middleware.py
nanaian: Perhaps we could add a custom `X-No-Cookie: yes` header to all SSR API requests and detect that way?

ethteck: yeah that's fine, I just didn't know how to modify the header but if you help me figure it out, I agree that's better

backend/coreapp/serializers.py
ethteck: I believe it's just the primary key atm, but we can increase the depth as needed

nanaian: Does this not break the frontend?
Also I'd rather use API URLs rather than primary keys, it's more RESTy - see #272

PR #289, opened 2022-01-04
Remove msgstyle gcc

EpochFlame:
the flag obfuscates errors; we've removed it


nanaian:
I was playing around with this flag today and noted it would be better to use the default msgstyle for a better #44 implementation, +1 to removing it

nanaian (approved):

PR #290, opened 2022-01-05
Clean compiler loading logs and allow multiple compilers per dir

ethteck:
Expands the yaml parser to allow for a list format wherein multiple compilers can be specified for the same base directory
Adds old_agbcc as a compiler
Reduces verbosity of some of the logs for loading compilers

### DIFF
--- a/backend/coreapp/compiler_wrapper.py
+++ b/backend/coreapp/compiler_wrapper.py
@@ -12,11 +12,13 @@ import os
 from pathlib import Path
 import subprocess
 from dataclasses import dataclass
-from platform import uname
+from platform import platform, uname
 
 logger = logging.getLogger(__name__)
 
 DiffResult = Dict[str, Any]
+
+CONFIG_JSON = "config.json"
 
 PATH: str
 if settings.USE_SANDBOX_JAIL:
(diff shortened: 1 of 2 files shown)


backend/coreapp/compiler_wrapper.py
```diff
@@ -31,34 +33,48 @@
...
-                    binaries = (x for x in binaries_path.glob("*") if x.name != config_json)
-                    if next(binaries, None) != None:
-                        logger.debug(f"Enabling {compiler_id}.")
-                        ret[compiler_id] = config
-                    else:
-                        logger.debug(f"No binaries for {compiler_id}, ignoring.")
-                else:
-                    logger.warning(f"Error: {compiler_id} {config_json} is missing 'cc' and/or 'platform' field(s), skipping.")
+                if isinstance(config, dict):
+                    load_compiler(compiler_dir_name, config.get("cc"), config.get("platform"), ret)
```
mkst, line 71: not a huge fan of mutating the dictionary being passed in - prefer the entry be returned and added to the `ret` dict
```py
cfg = load_compiler(foo, bar, baz)
if cfg is not None:
  ret.update(cfg)
```

ethteck: gonna change it

mkst (approved):

PR #292, opened 2022-01-05
Projects MVP

### LINKED ISSUE #82: Project importing
image

nanaian:
Fixes #82

TODO:
- [x] A way for project maintainers to add functions other than the admin panel, or syncing
- [x] Detect match in fork/parent tree and show a banner if a scratch is matched elsewhere (`Scratch.objects.filter(score=0, target_assembly=self.target_assembly)` probably)
- [x] Pull repos in another thread
- [x] Add /:project/:function page
- [x] ~~Don't use subprocess for git~~
- [x] Backend tests (how do we mock the github api?) - I need help with this
- [x] Replace `tools/decompme` with config file or database
- [x] papermario context fixes
- [x] Hide "10 attempts" placeholder
- [x] Rename exception classes
- [x] Squash migrations

Future stuff:
- [ ] Replace icon_url with FileField (static uploads)?
- [x] Attempts/forks tree
- [x] Replace "start attempt" with "continue attempt" if you have one
- [x] Store filename, complexity, asm length, etc. in ProjectFunction & display
- [ ] Allow pull with a token query parameter (for webhooks)
- [ ] Store target asm per-ProjectFunction instead of per-scratch

Database layout
image


backend/coreapp/admin.py
```diff
@@ -1,10 +1,32 @@
...
+from .models import Profile, Assembly, Asm, Scratch, Project, ProjectFunction
+from .github import GitHubUser, GitHubRepo
+
+class GitHubRepoAdmin(admin.ModelAdmin):
+    actions = ["pull", "reclone"]
+
+    @admin.action(description="Pull selected repos from GitHub")
+    def pull(self, request, queryset):
+        for repo in queryset.all():
+            repo.is_pulling = False
```
ethteck, line 14: Why false here and below in reclone?

backend/coreapp/models.py
```diff
@@ -60,11 +78,12 @@ class Scratch(models.Model):
     diff_label = models.CharField(max_length=512, blank=True, null=True)
     score = models.IntegerField(default=-1)
     max_score = models.IntegerField(default=-1)
-    parent = models.ForeignKey("self", null=True, blank=True, on_delete=models.CASCADE)
-    owner = models.ForeignKey(Profile, null=True, blank=True, on_delete=models.SET_NULL)
+    parent = models.ForeignKey("self", null=True, blank=True, on_delete=models.CASCADE) # XXX: should be SET_NULL?
```
ethteck, line 81: agree

backend/coreapp/models.py
```diff
@@ -60,11 +78,12 @@ class Scratch(models.Model):
...
+    owner = models.ForeignKey(Profile, null=True, blank=True, on_delete=models.SET_NULL) # XXX: should be CASCADE?
```
ethteck, line 82: If the owner is deleted, I think we can just set owner to null as well...

backend/coreapp/models.py
```diff
@@ -83,5 +102,32 @@ def __hash__(self):
             self.owner,
         ))
 
+    def save(self, *args, **kwargs):
+        # Scratches cannot be owned if they are a root function in a project
+        if self.owner is not None and ProjectFunction.objects.filter(scratch=self).exists():
+            self.owner = None
```
ethteck, line 108: I'm a bit confused about this - if a person tries to claim ownership of a base scratch for a project function, I think that should be an error, not just reject the new owner

ethteck (dismissed):
The mypy issues make this difficult to review, as it looks like some code in the models / serializers has references to nonexistant fields and such. Could you resolve those issues?

backend/coreapp/github.py
```diff
@@ -133,3 +135,93 @@ def login(request: Request, oauth_code: str) -> "GitHubUser":
...
+
+        self.is_pulling = True
+        self.save()
+
+        try:
+            repo_dir = self.get_dir()
+            remote_url = f"https://github.com/{self.owner}/{self.repo}"
+
+            if repo_dir.exists():
+                subprocess.run(["git", "remote", "set-url", "origin", remote_url], cwd=repo_dir)
```
mkst, line 164: might want to look at something like https://www.pygit2.org/ if this moves out of MVP

backend/coreapp/admin.py
nanaian: It forces a pull even if the database reckons its already doing one

backend/coreapp/models.py
nanaian: It does reject -- see the `claim` method on ScratchViewSet.

The reason this is here is because adding a project function in the admin interface or w/e should also set owner to None

backend/coreapp/github.py
nanaian: Personally I'm not entirely certain why this would be better than just using the shell git; we don't need to do very much with git

ethteck: because if we do it in python, we don't have to deal with the messiness of external processes and interpreting error codes / stderr / stdout. we just have a simple python object to work with. in general I think avoiding raw process calls from python is cleaner and better design

backend/coreapp/models.py
ethteck: I didn't mean the claim endpoint , I meant just adding a non-null owner for a scratch that's a base function. I think in all cases, that should be an error and not just silently ignore the specified owner

backend/coreapp/admin.py
ethteck: could you explain the scenario in which this might happen? I would imagine that if the database thinks there is one happening, we should leave it alone until it's done. 

On the other hand, if there *is* a pull happening in fact and we go to do it again, that could corrupt the state of the repo

nanaian: Good point

backend/coreapp/models.py
nanaian: So you think this should raise an error rather than just set it to None?

ethteck: I'm thinking about it, and I think this is fine but we should probably also be catching this in the viewset function for create and just not allow the owner to be explicitly set on create...? since the backend sets the owner itself via the profile, right?

backend/coreapp/models.py
nanaian: I guess I was thinking that deleting your account would also imply you want all data deleted

backend/coreapp/views/scratch.py
```diff
@@ -22,13 +22,18 @@
...
-from ..models import Asm, Scratch
+from ..models import Asm, Project, ProjectFunction, Scratch
 from ..middleware import Request
 from ..decorators.django import condition
 from ..serializers import TerseScratchSerializer, ScratchCreateSerializer, ScratchSerializer
+from ..github import GitHubRepo, GitHubRepoBusy
 
 logger = logging.getLogger(__name__)
 
+class ProjectNotMaintainer(APIException):
```
ethteck, line 33: all exception classes should end in Exception imo

ethteck (commented):
Found this minor thing as well as the larger topic I raised in #development (along with existing todos and the merge conflicts)

backend/coreapp/views/scratch.py
ethteck: also I'd call this NotProjectMaintainerException cause "Project not maintainer" sounds like "the project is not the maintainer" to me

backend/coreapp/github.py
nanaian: image

ethteck (approved):
lgtm! great work on this :) exciting shtoof to say the least. we can merge once it's not visible from the front page

PR #296, opened 2022-01-06
Fix preferences dialog on Safari

### LINKED ISSUE #284: Use custom modal dialog instead of using HTML <dialog>
`<dialog>` is really buggy on some browsers (read: older Safari) and its getting annoying to manually fix issues; we should just use a react modal library

nanaian:
Fixes #284


ethteck:
image
other than me wishing it wasn't this blurry in the background on chrome, lgtm!

Issue #141, opened 2021-10-05
Add "decompile" button to scratch page


ethteck:
Users should be able to "reset" their scratch by re-invoking m2c and getting back the original decompilation. 

Later, we will have options in the scratch that affect m2c's behavior, so this functionality will be for more than just resetting the current c code.

**Backend**: Expose a new endpoint that takes a scratch ID (and optionally context that can differ from that of the scratch) and returns m2c output
**Frontend**: Add a button that, when pressed, calls the endpoint and either

* displays (via pop-out window?) the decompiled code which can then be copy/pasted or just viewed for reference
*or*
* replaces the c code with the endpoint's result

ethteck:
(in #360)

PR #302, opened 2022-01-07
Decompile endpoint

ethteck:
Add decompile endpoint, tests, refactor tests a bit
backend part of #141


mkst (approved):

backend/coreapp/decompiler_wrapper.py
```diff
@@ -0,0 +1,28 @@
+import logging
+
+from coreapp.compiler_wrapper import CompilerWrapper
+from coreapp.m2c_wrapper import M2CError, M2CWrapper
+
+logger = logging.getLogger(__name__)
+
+
```
nanaian, line 8:
```suggestion
```

ethteck: Classes should have two empty lines before them, per pep8

nanaian: <small>pep8 is gross</small>

We should introduce a code formatter like I have with the TypeScript code like `black` I think

PR #309, opened 2022-01-11
Assemble NDS asm in thumb mode

abaresk:
The majority of arm9 code was compiled in thumb mode (at least for the Pokémon projects). This change adds a flag to assemble asm in thumb mode, as well. This fixes the issue where instruction widths were not matching for NDS scratches.

See example scratch with fix applied:

Screen Shot 2022-01-11 at 3 33 49 PM


mkst (approved):

ethteck:
Thanks a bunch for this PR!

PR #333, opened 2022-02-02
Fix #332 - about page broken for forks

### LINKED ISSUE #332: About page broken for forks
Parent field needs to return a url instead of a slug

ethteck:


backend/coreapp/serializers.py
```diff
@@ -99,6 +97,7 @@ class ScratchSerializer(serializers.ModelSerializer["Scratch"]):
     slug = serializers.SlugField(read_only=True)
     url = UrlField()
     html_url = HtmlUrlField()
+    parent = HtmlUrlField()
```
nanaian, line 100: Pretty sure this should be UrlField, not HtmlUrlField

ethteck: why? it wants to be a link the user clicks, not the API link

nanaian: No - see `AboutScratch.tsx`; the fork link is `<ScratchLink url={scratch.parent} />`, and ScratchLink makes an API request to the given `url` parameter to gather information about the scratch (such as its name and html_url).

ethteck: that's odd, cause it worked for me locally

nanaian (approved):

PR #334, opened 2022-02-03
Add `clang` compiler and `switch` platform

### LINKED ISSUE #75: Add clang (Switch) compiler support

MonsterDruide1:
This PR adds the Nintendo Switch as decompiler platform, along with two versions of the `clang` compiler and respective presets. It will fix #75 when merged.

First off, `clang-4.0.1` is pretty straight-forward. It's used by Breath of the Wild, so that one has been added as a preset to use too.

There is `clang-3.9.1` as well, as used by Super Mario Odyssey, which is missing the `lld` linker by default. To work around this issue, it is using the linker of `clang-4.0.1`, so those two compiler versions have to be installed to the same directory to make the relative path of `-fuse-ld=` work.

The language standards will probably go unused for the most part, but for completion I've added all `C++`-compatible versions here.

`TODO` before merging:
- [x] Include proper logo for the platform
- [x] A bit more testing, maybe by someone who knows `decomp.me` and its possible flaws better than me


nanaian (changes requested):
Looks good so far!
I'd ask that you ensure that both those `config.json` files for the compilers are terminated with newlines

ethteck:
As discussed in discord, let's copy ld from one compiler dir to the other just to make it a little less scary when invoking the compiler and having to go backwards in the directory tree. 

Also, if you could move the clang downloads into their own script (will need to modify the docker stuff for this, let us know if you don't find what I mean by that) and have it download [4.0.1](https://releases.llvm.org/4.0.1/clang+llvm-4.0.1-x86_64-apple-darwin.tar.xz) if the machine is running mac like we do for ido, that'd be greatly appreciated

Other than that, looks great! I'm working on fixing the issue with reviewdog, which is unrelated to your PR

ethteck:
another thing I just realized: for ido and gcc we do ido7.1 (no dash), and I see you added a dash between clang and the version. It doesn't look bad in isolation, but it bothers me that we're not consistent...same with mwcc, but for that we have an underscore lol. anyone have thoughts on this? 

Personally, I think we can tackle mwcc another day, but we should remove the dash from clang to be consistent with everything else

MonsterDruide1:
I'm not sure whether there is a "proper" way to write the clang versioning, but I've only seen it in this structure so far. I have to say though, my knowledge is limited to BotW on that end, so maybe @leoetlino knows more about it.

MonsterDruide1:
> Also, if you could move the clang downloads into their own script (will need to modify the docker stuff for this, let us know if you don't find what I mean by that) and have it download [4.0.1](https://releases.llvm.org/4.0.1/clang+llvm-4.0.1-x86_64-apple-darwin.tar.xz) if the machine is running mac like we do for ido, that'd be greatly appreciated
Should I also move the `apt-get` stuff specific for this compiler to a conditional thing, or should those (rather small) dependencies be installed either way?

ethteck:
I don't mind, but maybe @mkst does

mkst:
I'm easy, the other compilers (e.g. psx/wii) install (apt-get) deps separately, but as you say, they are only small...

MonsterDruide1:
> Also, if you could move the clang downloads into their own script (will need to modify the docker stuff for this, let us know if you don't find what I mean by that) and have it download [4.0.1](https://releases.llvm.org/4.0.1/clang+llvm-4.0.1-x86_64-apple-darwin.tar.xz) if the machine is running mac like we do for ido, that'd be greatly appreciated

Moving it into a separate script - done. About the mac thing tough, it will only work with `4.0.1`. `3.9.1` does not have a pre-built MacOS build, so if no one creates such a build manually we can't supply this compiler. Without it, `Super Mario Odyssey`s preset won't work though... will it be automatically hidden when the required compiler is not supported?

On a different note, can I merge the first and last commit (icon)? Otherwise it will have a "broken", stolen image for a few commits in the history...

ethteck:
> will it be automatically hidden when the required compiler is not supported?

yep! no issues here. and it's fine to exclude 3.9.1 on mac. just nice to have the one version since it's already there

> can I merge the first and last commit (icon)?

yep, go for it! We squash all PRs anyway but feel free

MonsterDruide1:
MacOS is implemented for `4.0.1` now. I have no way to check whether the commands are correct and if the file setup actually works, so someone else has to test this part.

ethteck (approved):
I know that this is in draft stage, but I went ahead and tested it and fixed some issues I found with the shell script and simplified it a bit. I tested the new platform and compilers, and they work great. Thanks so much for your work on this!

MonsterDruide1:
It is in draft state because I wanted someone to do more testing on this (and as a proper logo was missing at first). Other than that, it's finished from my end, so feel free to put it in "normal mode" when you're ready to merge.

nanaian:
On Mac the compilers appear to install correctly, but it looks like the downloaded binutils isnt being used:
image

nanaian (approved):

PR #343, opened 2022-02-05
Use objdump --disassemble to avoid diffing past function end (when possible)

leoetlino:
...and only when a diff_label was specified by the user.

For some modern platforms it's impossible to force the compiler to only generate code for a single function *even if* the source code and context tabs only contain one function: the compiler may need to generate e.g. PLT stubs for position-independent code.

### DIFF
--- a/backend/coreapp/asm_diff_wrapper.py
+++ b/backend/coreapp/asm_diff_wrapper.py
@@ -45,6 +45,9 @@ class AsmDifferWrapper:
 
     @staticmethod
     def get_objdump_target_function_flags(sandbox: Sandbox, target_path, platform: str, label: str) -> List[str]:
+        if compiler_wrapper.supports_objdump_disassemble(platform):
+            return [f"--disassemble={label}"]
+
         nm_command = compiler_wrapper.get_nm_command(platform)
         if not nm_command:
             raise NmError(f"No nm command for {platform}")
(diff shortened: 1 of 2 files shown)


backend/coreapp/asm_diff_wrapper.py
```diff
@@ -43,6 +43,38 @@ def create_config(arch: asm_differ.ArchSettings) -> asm_differ.Config:
             algorithm="levenshtein",
         )
 
+    @staticmethod
+    def get_objdump_target_function_flags(sandbox: Sandbox, target_path, platform: str, label: str) -> List[str]:
+        if compiler_wrapper.supports_objdump_disassemble(platform):
```
ethteck, line 48: we should add `and label` to this check I think, since it might not always be non-empty, even if the platform supports it

ethteck (approved):
Looks good! Just had this one thing

backend/coreapp/asm_diff_wrapper.py
leoetlino: ah, `label` can be empty but not None, right... thoughts on moving the `if label` check into this function?

ethteck: Yeah, that's what I'm advocating for

ethteck: ohhh I see, it was outside of this function. my b. yeah just putting it here is good

PR #349, opened 2022-02-10
Add GCC 2.6.3 for PS1

Xeeynamo:
Some PlayStation 1 games released between 1995 and 1997 uses GCC 2.6.x. As this compiler is not yet installed in the repo, this pull request aims to add its support.

Note that GCC 2.6.3 will always be downloaded from the tag `release` from [decompals' gcc repo](https://github.com/decompals/old-gcc/releases/tag/release).

I am marking the PR as draft as I am not currently able to spin-up an instance of decomp.me locally, but I also want to get feedback from the proposed change.

### DIFF
--- a/backend/compilers/download.sh
+++ b/backend/compilers/download.sh
@@ -40,6 +40,15 @@ if [[ "$uname" != "Darwin" ]]; then
     find . -name "CC1PSX.EXE" | xargs chmod +x
 fi
 
+# GCC mipsel
+if [[ "$uname" != "Darwin" ]]; then
+    curl -L "https://github.com/decompals/old-gcc/releases/download/release/gcc-2.6.3.zip" && unzip gcc-2.6.3.zip -d "$compiler_dir/gcc2.6.3-mipsel" && gcc-2.6.3.zip
+    find . -name "cc1" | xargs chmod +x
+    find . -name "cpp" | xargs chmod +x
+    find . -name "g++" | xargs chmod +x
+    find . -name "gcc" | xargs chmod +x
+fi
+
 # agbcc (gba)
 if [[ "$uname" != "Darwin" ]]; then
     mkdir -p "$compiler_dir/agbcc"
--- a/backend/compilers/gcc2.6.3-mipsel/config.json
+++ b/backend/compilers/gcc2.6.3-mipsel/config.json
@@ -1,0 +1,4 @@
+{
+    "platform": "ps1",
+    "cc": "\"${COMPILER_DIR}\"/gcc -G0 -c -B \"${COMPILER_DIR}\"/ $COMPILER_FLAGS \"$INPUT\" -o \"$OUTPUT\""
+}
--- a/frontend/src/components/compiler/compiler_settings/gcc.tsx
+++ b/frontend/src/components/compiler/compiler_settings/gcc.tsx
@@ -1,0 +1,15 @@
+import { Checkbox, FlagSet, FlagOption } from "../CompilerOpts"
+
+export function CommonGccFlags() {
+    return <>
+        <FlagSet name="Optimization level">
+            <FlagOption flag="-O0" description="No optimization" />
+            <FlagOption flag="-O1" description="Some optimization" />
+            <FlagOption flag="-O2" description="Standard optimization" />
+        </FlagSet>
+
+        <Checkbox flag="-fforce-addr" description="Load pointers into registers before use" />
+
+        <Checkbox flag="-Wall" description="Enable all warning types" />
+    </>
+}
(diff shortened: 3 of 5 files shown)


backend/compilers/download.sh
```diff
@@ -40,6 +40,15 @@ if [[ "$uname" != "Darwin" ]]; then
     find . -name "CC1PSX.EXE" | xargs chmod +x
 fi
 
+# GCC mipsel
+if [[ "$uname" != "Darwin" ]]; then
```
mkst, line 44: Would be nicer if the package had them already executable 🙂

backend/compilers/gcc2.6.3-mipsel/config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "platform": "ps1",
+    "cc": "\"${COMPILER_DIR}\"/gcc -G0 -c -B \"${COMPILER_DIR}\"/ $COMPILER_FLAGS \"$INPUT\" -o \"$OUTPUT\""
```
mkst, line 3: `-G0` is exposed as an option, so hard-coding here means users can't try out e.g. -G4/_G8

frontend/src/components/compiler/compilers/gcc2.6.3-mipsel.tsx
```diff
@@ -0,0 +1,6 @@
+import { CommonGccFlags } from "../compiler_settings/gcc"
+
+export const name = "GCC 2.6.3"
+export const id = "gcc263"
```
mkst, line 4: Think this needs to align with the folder name, so `-mipsel`

backend/compilers/download.sh
```diff
@@ -40,6 +40,15 @@ if [[ "$uname" != "Darwin" ]]; then
...
+    curl -L "https://github.com/decompals/old-gcc/releases/download/release/gcc-2.6.3.zip" && unzip gcc-2.6.3.zip -d "$compiler_dir/gcc2.6.3-mipsel" && gcc-2.6.3.zip
+    find . -name "cc1" | xargs chmod +x
+    find . -name "cpp" | xargs chmod +x
```
mkst, line 47: Do you mean to `rm <zipname>` rather than trying to execute it at the end of this line?

frontend/src/components/compiler/compiler_settings/gcc.tsx
```diff
@@ -0,0 +1,15 @@
+import { Checkbox, FlagSet, FlagOption } from "../CompilerOpts"
```
ethteck, line 1: It's good to have a common gcc settings file like this, and we should probably use it in all the other gcc compilers too

backend/compilers/download.sh
ethteck: I think you mean a couple lines up, but yeah agreed looks like a typo

backend/compilers/download.sh
ethteck: yeah that'd save every consumer of this the trouble

ethteck (commented):
After these changes are made (if you don't mind incorporating those), I'll try testing locally

backend/compilers/download.sh
Xeeynamo: Yes it is definitively a typo!

nanaian (commented):
Nothing to add :)

Xeeynamo:
I was able to run it locally and I found few things to fix here and there. I marked the PR as ready to review.

It is a good idea to mark those zip archives from `old-gcc` as eXecute, but I am not a fan of creating a tight dependency across the two projects so I left the `chmod +x` if you do not mind it.

backend/compilers/gcc2.6.3-mipsel/config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "platform": "ps1",
+    "cc": "mips-linux-gnu-cpp -Wall -lang-c -gstabs \"$INPUT\" | \"${COMPILER_DIR}\"/cc1 -mips1 -mcpu=3000 -G0 -fno-builtin $COMPILER_FLAGS | mips-linux-gnu-as -march=r3000 -mtune=r3000 -no-pad-sections -O1 -o \"$OUTPUT\""
```
mkst, line 3: Think `-G0 -fno-builtin` should be exposed as flags rather than hardcoded

mkst: Also wonder if the `-O1` will become a sticking point. @ethteck we briefly talked about exposing assembler flags but it's nontrivial I guess..

ethteck: yeah it's a whole other field for the model + api so I agree it's useful but ...it work

ethteck:
Minor nit: could you make your new config file end in a new line?

ethteck:
Also see reviewdog feedback - it found two things I think? an extra semicolon somewhere and non-alphabetical import of something

ethteck:
oh...lame. okay no problem. I think it's a local setting I have configured, but we should be able to have a settings file for the project that enforces that.

Xeeynamo:
image
For the sake of transparency, I am currently getting this error only when creating a new scratch regardless the compiler used. I am rolling back commit by commit to see if is a problem from the PR or my local environment acting weird again lol

Issue #326, opened 2022-01-29
Scratch export doesn't work as expected


ethteck:
https://decomp.me/scratch/r5pf4

No clue on the scope, but it is probably related to the arch

Edit: nevermind, it's happening across the board. should be easy to fix, I think

nanaian:
I'm getting this on Chrome
![image](https://user-images.githubusercontent.com/9429556/151777834-82211889-c188-482f-8988-92c51c6aa72e.png)

nanaian:
Any idea in which commit this was introduced?

ethteck:
no clue :/ I'll try to reproduce locally

ethteck:
ah I got it. it's cause the url here is not a full url, it's the relative part
```
function exportScratchZip(scratch: api.Scratch) {
    const url = `${scratch.url}/export`
    const a = document.createElement("a")
    a.href = url
    a.download = scratch.name + ".zip"
    a.click()
}```

PR #351, opened 2022-02-10
frontend: Fix scratch exporting (#326)

MonsterDruide1:
Related: Issue [#326](https://github.com/decompme/decomp.me/issues/326)

This fixes the exporting of scratches as `zip` files from the site.

Due to earlier changes, it seems like some endpoint for scratches was moved between frontend and backend. The `scratch.url` only contains the "ending" of the URL, so for example `/scratch/bvwj6`. It does *not* contain the URL/API http-endpoint, so the browser relates it to the frontend instead of the backend. Instead, the URL has to be resolved to point to the exporting part of the backend.

As the duplicate code would start to pile up for resolving this kind of "URL-ending" to a full URL, I moved that to a separate function that can be called for directly getting API data, or only for resolving the URL like in this case.

### DIFF
--- a/frontend/src/components/Scratch/ScratchToolbar.tsx
+++ b/frontend/src/components/Scratch/ScratchToolbar.tsx
@@ -29,7 +29,7 @@ function htmlTextOnly(html: string): string {
 }
 
 function exportScratchZip(scratch: api.Scratch) {
-    const url = `${scratch.url}/export`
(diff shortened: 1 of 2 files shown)


ethteck (approved):
Thanks so much for doing this. I'll leave it to alex to review the change, but let's merge it if he's cool with it.

nanaian (approved):

Issue #300, opened 2022-01-06
Improve automatically recompile on change feature


ethteck:
This feature seems a bit too aggressive as to when it chooses to send an outgoing compile request - it often does so while the user is in the middle of typing. It might be worth exposing a slider in the preferences that allows the user to set a millisecond delay between the last keystroke and an automatic recompilation.

There's also a bug where sometimes a response will come back while the user is still typing and the front-end doesn't seem to think it needs to send out a new request, so the user has to manually invoke a compilation.

MonsterDruide1:
I'm not able to reproduce the bug of simultaneous request/response. Any hints on what exactly causes this?

nanaian:
I'm pretty sure I fixed that bug a while back @MonsterDruide1

MonsterDruide1:
Okay, if that part got fixed, the mentioned PR can be marked as "linked" to close this issue on merge.

ethteck:
No, the bug hasn't been fixed. There are times where, if the site is waiting for an incoming compile result and the user types before the result comes back, the client doesn't send out another request. I don't have a super easy way to reproduce it

nanaian:
We can fix this if we wrote this hook ourselves:
https://github.com/decompme/decomp.me/blob/main/frontend/src/lib/api.ts#L399

MonsterDruide1:
I can reproduce this as suggested, using the "Latency"-emulation by Google Chrome. It's still an issue, and I'm working on resolving that. However, as I started a custom implementation of `DebouncedCallback`, it might still take a while before it's finished, so I think this has to go into a separate PR from #353 .

PR #353, opened 2022-02-10
frontend: Add customizable delay before compile

MonsterDruide1:
Related: Issue #300

This adds a slider and input field to the settings page, which allows to customize the delay between typing and re-compiling the file.

The input fields are configured like this: 
```html
min="50" max="2000" step="50"
```

I'm not sure this is actually a good range and should be kept like this, I'd like to hear opinions of others on this range. `50ms` works locally for me, but might nuke the server with too many requests?

The default value is kept at `500`, as before.

Preview: New setting for adjusting time before recompile


frontend/src/components/Scratch/ScratchPreferencesModal.tsx
```diff
@@ -22,6 +23,19 @@ function DiffPrefs() {
...
+                <input
+                    type="range"
+                    min="50" max="2000" step="50" value={autoRecompileDelay}
+                    onChange={evt => setAutoRecompileDelay(evt.target.value)}
+                />
+                <input className={styles.numberInput}
+                    type="number"
+                    min="50" max="2000" step="50" value={autoRecompileDelay}
+                    onChange={evt => setAutoRecompileDelay(evt.target.value)}
+                />
```
nanaian, line 36: I think it would be better to implement this as a ContentEditable so there isn't whitespace between the number and "ms". See the scratch name component for an example of this

nanaian (changes requested):

frontend/src/components/Scratch/ScratchPreferencesModal.tsx
MonsterDruide1: That way, there are no arrows for increasing/decreasing the value by clicking. Which one is preferred?

nanaian: I don't think the arrows are particularly needed since we have that range input next to it anyway

nanaian: We might be able to make our own fancy `TimePeriodInput` component that includes the "ms" prefix such that you can type "5s" and have it be understood to mean 5000 ms

frontend/src/components/Scratch/ScratchPreferencesModal.tsx
```diff
@@ -22,6 +23,19 @@ function DiffPrefs() {
                 />
                 Automatically compile on change
             </label>
+            <label className={styles.intPreference}>
+                <input
+                    type="range"
```
nanaian, line 28: Consider disabling these inputs (and greying out this `label`) while "Automatically compile on changed" is unchecked

JoshDuMan:
Hope this stuff can be fixed, very happy to see this option!

nanaian:
I don't think either of my requests are really needed to get this deployed, they're moreso polish things. If @MonsterDruide1 is OK with it I can probably make the necessary changes myself so this can be deployed alonside #292

frontend/src/components/Scratch/ScratchPreferencesModal.tsx
MonsterDruide1: I'm not sure how quick or polished you want this to be merged, but I pushed the changes introducing a `TimePeriodInput`-component. It does not support different time units yet, but it's a `ContentEditable`.

frontend/src/components/Scratch/ScratchPreferencesModal.tsx
MonsterDruide1: Done. It now looks like this when disabled:
image

enabled:
image

frontend/src/components/Scratch/ScratchPreferencesModal.tsx
nanaian: Awesome, looks good. In retrospect I don't think we need to include the suffix parsing thing, it's a bit overkill unless we decide we want longer time period inputs elsewhere in decomp.me

nanaian (approved):

ethteck (approved):

PR #356, opened 2022-02-17
Use mock/patch to write GitHubRepo tests

matt-kempster:


backend/coreapp/tests.py
```diff
@@ -683,15 +684,31 @@ def create_test_project(self, slug: str = "test") -> Project:
...
         project = self.create_test_project()
-
         project.repo.pull()
-        self.assertTrue(project.repo.get_dir().is_dir())
 
+        mock_subprocess.assert_called_once()
+        self.assertListEqual(
+            mock_subprocess.calls[0][:2],
+            ["git", "clone", f"https://github.com/{owner}/{repo}"]
+        )
```
nanaian, line 700: I think this should look at any of the calls, not just the 0th

matt-kempster: There's only one call (a fact that was asserted on the previous line)

nanaian: oop, my bad :+1:

nanaian (approved):

PR #357, opened 2022-02-17
Make max_score populate even if the compilation fails

ethteck:
Also do some refactoring, make it so scratches don't update when we update their score/max_score

fixes #357


nanaian (changes requested):
LGTM! 📦 

> make it so scratches don't update when we update their score/max_score

Could you add a test that verifies this?

nanaian (approved):

PR #366, opened 2022-02-18
Poetry

ethteck:


README.md
```diff
@@ -237,8 +236,8 @@ mypy
 
 If you modify any database models (`models.py`), you'll need to run the following to update the database:
 ```shell
-python manage.py makemigrations
-python manage.py migrate
+poetry run python manage.py makemigrations
+poetry run python manage.py migrate
```
nanaian, line 240: Shouldn't this be
```suggestion
poetry run manage.py makemigrations
poetry run manage.py migrate
```

nanaian (changes requested):

README.md
ethteck: no, per their docs you run stuff normally with `python` but you just prefix commands with `poetry run`

nanaian (approved):

backend/Dockerfile
```diff
@@ -48,9 +49,25 @@ RUN apt-get -y update && apt-get install -y \
...
+    && chown -R user:user /sandbox \
+    && mkdir -p "${WINEPREFIX}" \
+    && chown user:user "${WINEPREFIX}"
+
+COPY pyproject.toml /backend/pyproject.toml
+COPY poetry.lock /backend/poetry.lock
+
+# Install poetry and dependencies
+USER user
+RUN curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/install-poetry.py | python -
```
mkst, line 67: Where are the "poetry" deps installed to? the `/home/user/.<something>`...

trying to decide whether this should all be done in the docker image (as it was with requirements.txt) or as part of entrypoint.sh (as the frontend entrypoint does -- but node_modules are downloaded into /frontend so persist)

ethteck: I actually don't know... shorry

.github/workflows/pr.yml
```diff
@@ -129,7 +114,9 @@ jobs:
             -e COMPILER_BASE_PATH=/compilers \
             -e WINEPREFIX=/tmp/wine \
             -e LOCAL_FILE_DIR=/local_files \
-            decompme_backend -c 'cd /decomp.me && python backend/manage.py test backend'
+            decompme_backend -c 'cd /decomp.me/backend && \
+            /home/user/.local/bin/poetry install && \
```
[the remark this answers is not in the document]
ethteck: Yeah I'm confused because it should work given that we install them in the Dockerfile, but it doesn't...so I guess we should remove them from the Dockerfile? But that's annoying. I was hoping you could help me figure out why it doesn't work to just execute poetry (why it can't find the deps)

PR #376, opened 2022-02-18
Use psyq assembler

mkst:
removing "psyq4.3" as there is no psyq cc + assembler combo


nanaian (dismissed):
This needs a database migration that converts scratches with `compiler=psyq4.3` to `compiler=gcc2.7.2-psyq`

backend/coreapp/migrations/0017_rename_psyq43_compiler.py
```diff
@@ -0,0 +1,27 @@
...
+
+def rename_psyq43(apps, schema_editor):
+    """
+    Restore forks' projects
+    """
+    Scratch = apps.get_model("coreapp", "Scratch")
+    for row in Scratch.objects.all():
+        if row.compiler == "psyq4.3":
+            row.compiler = "gcc2.7.2-psyq"
+        row.save()  # type: ignore
```
ethteck, line 13: Bump row.save up into the if (no need to save unless we change the value), remove #type: ignore (my mistake to put it in the previous migration). 

Also you should supply change_fields like we do in the latest version of 0016. Pull master into this to see what I mean

ethteck (changes requested):

backend/coreapp/migrations/0017_rename_psyq43_compiler.py
mkst: I thought I might only need to save the changed rows, but `0013_scratch_name_default` was saving every row, and so I just copied that :)

ethteck (approved):

PR #396, opened 2022-02-27
Compiler/platform refactor

ethteck:
Refactors compiler / platform stuff into a more OOP design
Downloading happens in compilers/download.py, and no files are checked into the repo under compilers/ other than this script. 
Downloads are stored in compilers/download_cache 
Compiler configurations are stored in the Compiler instance itself rather than in a config file on disk

~~TODO:~~
- [x] Modify dockerfile to use the new download.py script
- [x] Add PS1 compilers to download.py
- [x] Confirm new gcc272kmc compiler works
- [x] Add migration from gcc27kmc to gcc272kmc
- [x] Add migration from mwcc_247_108_tp to mwcc_247_108, since the former has been removed

Flags:
- [x] Store flags into Compiler
- [x] Return flags as part of /compilers endpoint
- [x] Change frontend to accommodate new flag format
- [x] Put flag strings into i18n file (set up framework for this)

Presets:
- [x] Store presets in new Preset class
- [x] Return presets as part of /presets endpoint
- [x] Change frontend to use presets from new endpoint, remove stored ones


frontend/src/components/compiler/CompilerOpts.tsx
```diff
@@ -52,7 +56,7 @@ export function FlagOption({ flag, description }: { flag: string, description?:
         value={flag}
         selected={checkFlag(flag)}
     >
-        {flag} {description && `(${description})`}
+        {flag} {description && description !== "NULL" && `(${description})`}
```
ethteck, line 59: we can't return the empty string by default or `null`, so this is what I settled on. thoughts?

nanaian (approved):

frontend/src/components/compiler/CompilerOpts.tsx
nanaian: is fine with me

PR #404, opened 2022-03-13
[WIP, testers needed] implement gcc 4.0.1 cross compiler from amd64 linux to ppc mac os x

ChrisNonyminus:
This adds a cross compiler from amd64 Linux to a PowerPC-based Mac OS X target, for decompiling some PowerPC-era Mac OS X games.
There may be some jank, and the code's a little spaghetti-ish, so testers and reviews are needed.

Note: I suck at rebasing, so the rebase commit on this branch accidentally undid some changes. That shouldn't be too hard to fix though.

### DIFF
--- a/backend/compilers/download.py
+++ b/backend/compilers/download.py
@@ -183,6 +183,32 @@ def download_gba():
         "agbccpp",
     )
 
+def download_ppc_darwin():
+    if host_os != LINUX:
+        return
+    download_zip(
+        url="https://github.com/ChrisNonyminus/powerpc-darwin-cross/releases/download/initial/powerpc-darwin-cross.new.zip",
+        dl_name="powerpc-darwin-cross.zip",
+        dest_name="powerpc-darwin-cross",
+        create_subdir=True,
+    )
+    download_zip(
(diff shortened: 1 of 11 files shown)


ethteck:
Thanks for your PR! a few initial thoughts: 
* this PR involves adding a new platform to the site. In order to do that, need to find/create a logo svg for it
* I'm still kind of unsure as to the use of this on the site before there are any projects that actually use this compiler. I think a project that uses the compiler and is actively being worked on should be kinda a prerequisite to adding a new platform / compiler. I would make exceptions for things like official modern gcc/clang releases, but this kind of thing is very niche, and there aren't any active projects using it yet
* aside from the cleanup required, there are some mypy/black issues as you can see in the checks at the bottom of the PR window - let us know if you need help getting these to pass

ethteck:
I don't think "Darwin" is right either, as modern x86 apple computers are also known as "darwin". so we probably would want to call it "PowerPC Mac" or something like that

ChrisNonyminus:
Ok, understood.

ethteck:
closing for now as discussed, but let's re-open this once you have a project going!

PR #416, opened 2022-03-30
Highlight assembly line on source selected line change

hatal175:
This could really use a good code review from someone who knows react.

### DIFF
--- a/frontend/src/components/Diff/Diff.module.scss
+++ b/frontend/src/components/Diff/Diff.module.scss
@@ -27,12 +27,10 @@
 .row {
     padding: 0 4px;
     height: 1.25em;
-}
 
-.hoveredSourceRow {
-    padding: 0 4px;
-    height: 1.25em;
-    background-color: var(--a150);
+    &.highlight {
+        background-color: var(--a150);
+    }
 }
 
 .header {
--- a/frontend/src/components/Diff/Diff.module.scss
+++ b/frontend/src/components/Diff/Diff.module.scss
@@ -29,6 +29,12 @@
     height: 1.25em;
 }
 
+.hoveredSourceRow {
+    padding: 0 4px;
+    height: 1.25em;
+    background-color: var(--a150);
+}
+
 .header {
     display: flex;
     align-items: center;
--- a/frontend/src/components/Diff/Diff.module.scss
+++ b/frontend/src/components/Diff/Diff.module.scss
@@ -29,7 +29,7 @@
     height: 1.25em;
 
     &.highlight {
-        background-color: var(--a150);
+        background-color: rgba(255, 255, 255, 0.1);
     }
 }
 
--- a/frontend/src/components/Diff/Diff.tsx
+++ b/frontend/src/components/Diff/Diff.tsx
@@ -37,7 +37,10 @@ function DiffColumn({ diff, prop, header, className, hoveredSourceLine }: {
         </div>
         <div className={styles.body}>
             {diff?.rows?.map?.((row, i) => (
-                <div key={i} className={(typeof row[prop]?.src_line != "undefined" && row[prop]?.src_line == hoveredSourceLine) ? styles.hoveredSourceRow : styles.row}>
+                <div key={i} className={classNames({
+                    [styles.row]: true,
+                    [styles.highlight]: (typeof row[prop]?.src_line != "undefined" && row[prop]?.src_line == hoveredSourceLine),
+                })}>
                     {typeof row[prop]?.src_line != "undefined" && <span className={styles.lineNumber}>{row[prop].src_line}</span>}
                     {row[prop] && <FormatDiffText texts={row[prop].text} />}
                 </div>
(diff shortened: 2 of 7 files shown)


frontend/src/components/Editor/CodeMirror.tsx
```diff
@@ -78,5 +85,25 @@ export default function CodeMirror({ value, onChange, className, viewRef: viewRe
         }
     }, [value])
 
-    return <div ref={el} className={className} style={{ fontSize: "0.8em" }} />
+    const debounceHover = debounce(50, false, (event: MouseEvent) => {
```
nanaian, line 88:
```suggestion
    const debouncedOnMouseMove = debounce(50, false, (event: MouseEvent) => {
```

frontend/src/components/Editor/CodeMirror.tsx
```diff
@@ -27,6 +29,11 @@ export default function CodeMirror({ value, onChange, className, viewRef: viewRe
     const extensionsRef = useRef(extensions)
     extensionsRef.current = extensions
 
+    const hoveredSourceLineref = useRef<number>()
```
nanaian, line 32:
```suggestion
    const hoveredSourceLineRef = useRef<number>()
```

frontend/src/components/Editor/CodeMirror.tsx
```diff
@@ -2,18 +2,20 @@ import { MutableRefObject, useEffect, useRef } from "react"
...
 import { Extension, EditorState } from "@codemirror/state"
 import { EditorView } from "@codemirror/view"
+import { debounce } from "throttle-debounce"
 
 import { materialPalenight } from "../../lib/themes/dark"
 
 export interface Props {
     value: string
     onChange?: (value: string) => void
+    OnHoverSourceChange?: (value: number | null) => void
```
nanaian, line 12:
```suggestion
    onHoveredSourceLineChange?: (value: number | null) => void
```

frontend/src/components/Diff/Diff.tsx
```diff
@@ -24,19 +24,20 @@ function FormatDiffText({ texts }: { texts: api.DiffText[] }) {
...
+    hoveredSourceLine?: number | null
 }) {
     return <resizer.Section className={classNames(styles.column, className)} minSize={100}>
         <div className={classNames(styles.row, styles.header)}>
             {header}
         </div>
         <div className={styles.body}>
             {diff?.rows?.map?.((row, i) => (
-                <div key={i} className={styles.row}>
+                <div key={i} className={(typeof row[prop]?.src_line != "undefined" && row[prop]?.src_line == hoveredSourceLine) ? styles.hoveredSourceRow : styles.row}>
```
nanaian, line 40:
```suggestion
                <div
                    key={i}
                    className={classNames({
                        [styles.row]: true,
                        [styles.highlight]: typeof hoveredSourceLine == "number" && row[prop]?.src_line == hoveredSourceLine}
                    })}
                >
```

frontend/src/components/Diff/Diff.module.scss
```diff
@@ -29,6 +29,12 @@
     height: 1.25em;
 }
 
+.hoveredSourceRow {
+    padding: 0 4px;
+    height: 1.25em;
+    background-color: var(--a150);
```
nanaian, line 35:
```suggestion

    &.highlight {
        background-color: var(--a150);
    }
```

frontend/src/lib/api.ts
```diff
@@ -192,6 +192,7 @@ export interface Scratch extends TerseScratch {
     context: string
     diff_label: string
     parent: string | null
+    hoveredSourceLine: number | null
```
nanaian, line 195: This piece of state should not be contained in Scratch.

nanaian (changes requested):
Nice PR! There are some issues with where state lives right now...

`hoveredSourceLine` state should not be contained in `Scratch`, for a couple reasons:
1. `Scratch` is directly reflective of what the API returns and expects
2. `hoveredSourceLine` is _editor state_ and not relevant to the rest of the scratch data (besides `source_code`)

Instead, you should do `const [hoveredSourceLine, setHoveredSourceLine] = useState<number | null>(null)` in the `<Scratch>` component (`frontend/src/components/Scratch/Scratch.tsx`) and pass down`setHoveredSourceLine` and/or `hoveredSourceLine` where required. For instance, CompilationPanel can just take `hoveredSourceLine` instead of the whole scratch, and it shouldn't need its own copy of that state or that useEffect handler.

If you need any help making these changes feel free to drop me a message on Discord :shipit:

frontend/src/components/Diff/Diff.tsx
nanaian: You may need to `import classNames from "classnames"`

hatal175: I haven't really changed anything here regarding classNames though...

nanaian (approved):

Issue #115, opened 2021-09-02
Rich diffs


nanaian:
With asm-differ's new [JSON output format](https://github.com/simonlindholm/asm-differ/commit/e055cb27603005338f32f0acc4f1eb1b2e487d0c), we can implement a richer diff interface that just the highlighted text we have right now.

Things I can think of that would be nice:
- [ ] Click a jump to scroll to it
- [ ] Click a register to find all uses of it
- [ ] Highlight the asm lines affected by the current line in the sourcecode editor

@simonlindholm @zbanks @ethteck perhaps you have more ideas?

nanaian:
#130 begins implementation of this :shipit:

1superchip:
- [ ] Scroll to the corresponding assembly from a line number in the source

bates64:
@1superchip please raise a separate issue for this. This issue is specifically for the diff UI element, so clicking on line numbers in CodeMirror would be out of scope even though it would scroll the diff

PR #435, opened 2022-04-09
React error fixes

simonlindholm:
Devtools console was showing some errors. No user impact, but this fixes them. (React devtools panel stills shows a warnings about missing translations for "compilers:mips_version.-mips1" and similar.)


frontend/src/components/Diff/Diff.tsx
```diff
@@ -21,13 +21,13 @@ const SelectedSourceLineContext = createContext<number | null>(null)
 
 function FormatDiffText({ texts }: { texts: api.DiffText[] }) {
     return <> {
-        texts.map(t => {
+        texts.map((t, i) => {
```
nanaian, line 24: [This is a poor fix](https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318) - I'd rather keep the warning rather than use the index as the key.
I'm not really sure what a suitable key _would_ be, though

frontend/src/components/Nav/Search.tsx
```diff
@@ -149,33 +133,31 @@ export default function Search({ className }: { className?: string }) {
...
-                            key={scratch.url}
+        {renderLayer(
+            <ul
+                {...getMenuProps(layerProps)}
+                className={classNames(verticalMenuStyles.menu, styles.results)}
+                style={{
+                    width: (triggerBounds ? triggerBounds.width : "unset"),
+                    ...layerProps.style,
+                }}
+                data-open={isOpen ? "true" : "false"}
```
nanaian, line 144: I'd rather use a class here than a data property. You can pass `{ [styles.open]: isOpen }` as an arg to the `classNames` call above to achieve this

nanaian (changes requested):

frontend/src/components/Diff/Diff.tsx
simonlindholm: I think this falls under the "static content" exception mentioned in that post:

> 1.    the list and items are static–they are not computed and do not change;
> 2.    the items in the list have no ids;
> 3.    the list is never reordered or filtered.
>
> When all of them are met, you may safely use the index as a key.

We shouldn't be updating individual lines in-place here, so it's fine to silence the warning.

nanaian (approved):

frontend/src/components/Diff/Diff.tsx
nanaian: That's true for now, my only worry would be with future #115 changes. But I guess this is alright for now, ty!

simonlindholm:
Hang on, I just found a bug in this, let's not merge it yet. It makes the dropdown menu invisible but still clickable.

nanaian (approved):

PR #452, opened 2022-05-20
Add MWCPPC (Mac OS 9 cross compiler ran on windows)

ChrisNonyminus:
Adds:
-The cross-compiler (target is Mac OS 9 (PowerPC), and host is Win32) and linker from Metrowerks CodeWarrior Pro 5 and 6.
-Some changes to work around the compilers' jank and work around the proprietary format used for the object files by dumping the assembly, converting it to a format binutils can assemble, and assembling a linux elf that can be dumped with objdump
-A preset for the aforementioned cross-compiler: The Sims 1, with flags "-lang=c++ -O3 -str pool -g"

As there is no version of doldisasm.py but for PEF files instead of DOLs (due to PEFs not only being proprietary, but also poorly documented and having no existing python library to use; I'm too lazy to implement an entire PEF reader in python myself as PEFs are really complex), scratches need to have the input asm be asm generated from IDA Pro, that has to be manually reformatted (a script to automate this should probably be made) into the GNU Assembler Syntax, in a way that can be assembled by binutils. [Here's an example of a reformatted assembly.](https://github.com/ChrisNonyminus/sims1_mac_decomp/blob/master/asm/split/CSimsApp.s)

Note: Asm diffing a scratch requires you to define a function in the scratch that has the same symbol as the diff label in the input asm.

Needs some cleaning, and a better solution for MWCPPC's jank and the object file format jank; we can't just have MWCPPC make an ELF; it doesn't do that as Mac OS 9 uses PEFs instead of ELFs.

### DIFF
--- a/backend/compilers/download.py
+++ b/backend/compilers/download.py
@@ -163,7 +163,7 @@ def download_zip(
 
 def download_codewarrior():
     download_zip(
-        url="https://github.com/ChrisNonyminus/sims1_mac_decomp/files/8735152/compilers.zip",
+        url="https://github.com/ChrisNonyminus/sims1_mac_decomp/files/8748510/compilers.zip",
         dl_name="codewarrior_compilers.zip",
         dest_name="codewarrior",
         create_subdir=True,
--- a/backend/compilers/download.py
+++ b/backend/compilers/download.py
@@ -177,7 +177,9 @@ def download_codewarrior():
         lowercase_lmgr = compiler_dir / ver / "lmgr8c.dll"
         if lowercase_lmgr.exists():
             shutil.move(lowercase_lmgr, compiler_dir / ver / "LMGR8C.dll")
-
+        # Set +x to allow WSL without wine
+        exe_path = compiler_dir / ver / "MWCPPC.exe"
+        exe_path.chmod(exe_path.stat().st_mode | stat.S_IEXEC)
         shutil.move(compiler_dir / ver, COMPILERS_DIR / "mwcppc_23")
 
 
--- a/backend/compilers/download.py
+++ b/backend/compilers/download.py
@@ -160,6 +160,7 @@ def download_zip(
         ):
             f.extract(member=file, path=dest_path)
 
+
 def download_codewarrior():
     download_zip(
         url="https://github.com/ChrisNonyminus/sims1_mac_decomp/files/8735152/compilers.zip",
@@ -170,11 +171,11 @@ def download_codewarrior():
     compiler_dir = COMPILERS_DIR / "codewarrior" / "compilers"
     lowercase_lmgr = compiler_dir / "Pro5" / "lmgr326b.dll"
     if lowercase_lmgr.exists():
-        shutil.move(lowercase_lmgr, compiler_dir / "Pro5"/ "LMGR326B.dll")
+        shutil.move(lowercase_lmgr, compiler_dir / "Pro5" / "LMGR326B.dll")
 
     lowercase_lmgr = compiler_dir / "Pro5" / "lmgr8c.dll"
     if lowercase_lmgr.exists():
-        shutil.move(lowercase_lmgr, compiler_dir/ "Pro5" / "LMGR8C.dll")
+        shutil.move(lowercase_lmgr, compiler_dir / "Pro5" / "LMGR8C.dll")
 
     # Set +x to allow WSL without wine
     exe_path = compiler_dir / "Pro5" / "MWCPPC.exe"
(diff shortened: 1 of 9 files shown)


backend/coreapp/views/scratch.py
```diff
@@ -62,6 +62,7 @@ def compile_scratch(scratch: Scratch) -> CompilationResult:
         scratch.compiler_flags,
         scratch.source_code,
         scratch.context,
+        scratch.diff_label,
```
mkst, line 65: Is this an artifact of the previous merge mess?

ChrisNonyminus: No, this is one of the changes I had to make for the compiler to work. The diff label has to be defined for the asm diffing process to work with the compiler.

backend/compilers/download.py
```diff
@@ -161,6 +161,43 @@ def download_zip(
...
+    if lowercase_lmgr.exists():
+        shutil.move(lowercase_lmgr, compiler_dir / "Pro6" / "LMGR326B.dll")
+
+    lowercase_lmgr = compiler_dir / "Pro6" / "lmgr8c.dll"
+    if lowercase_lmgr.exists():
+        shutil.move(lowercase_lmgr, compiler_dir / "Pro6" / "LMGR8C.dll")
+
+    # Set +x to allow WSL without wine
+    exe_path = compiler_dir / "Pro6" / "MWCPPC.exe"
+    exe_path.chmod(exe_path.stat().st_mode | stat.S_IEXEC)
```
ethteck, line 195: I'd prefer we do this in a loop like `for ver in ["Pro5", "Pro6"] rather than duplicating this code twice

backend/coreapp/compiler_wrapper.py
```diff
@@ -156,6 +156,71 @@ def compile_code(
...
+                    raise CompilationError(e.stderr)
+                except ValueError as e:
+                    # Shlex issue?
+                    logging.debug("Compilation failed: %s", e)
+                    raise CompilationError(str(e))
+
+            with open(sandbox.path / "c_asm.s", "r") as file:
+                for line in file:
+                    print("# %s" % line)
+
```
ethteck, line 223: I've mentioned before, but this shouldn't be necessary and causes a lot of tech debt wrt duplicated code that needs to be cleaned up now in two places.

At the very least, the common code should be abstracted. But I don't understand why we need two compiler invocations in the first place

backend/coreapp/compilers.py
```diff
@@ -34,6 +44,9 @@ class Compiler:
     is_gcc: ClassVar[bool] = False
     is_ido: ClassVar[bool] = False
     is_mwcc: ClassVar[bool] = False
+    is_mac_compiler: ClassVar[bool] = False
```
ethteck, line 47: `_compiler` is redundant here imo, unless you did that because it could imply the compiler's host machine is mac rather than the target?

backend/coreapp/compilers.py
```diff
@@ -256,6 +276,27 @@ def preset_from_name(name: str) -> Optional[Preset]:
     cc='cpp -P "$INPUT" | ${WINE} "${COMPILER_DIR}"/cc1n64.exe -quiet -G0 -mcpu=vr4300 -mips3 -mhard-float -meb ${COMPILER_FLAGS} -o "$OUTPUT".s && ${WINE} "${COMPILER_DIR}"/asn64.exe -q -G0 "$OUTPUT".s -o "$OUTPUT".obj && "${COMPILER_DIR}"/psyq-obj-parser "$OUTPUT".obj -o "$OUTPUT" -b -n',
 )
 
+# MACOS9
+MWCPPC_CC = 'printf "%s" "${COMPILER_FLAGS}" | xargs -x -- ${WINE} "${COMPILER_DIR}/MWCPPC.exe" -o object.o "${INPUT}"'
+MWCPPC_AS = 'printf "%s" "-dis -h -module ".${FUNCTION}" -nonames -nodata" | xargs -x -- ${WINE} "${COMPILER_DIR}/MWLinkPPC.exe" "${INPUT}" > ${COMPILER_DIR}/code.s'
+MWCPPC_CONVERT_GAS_SYNTAX = 'python3 ${COMPILER_DIR}/convert_gas_syntax.py "${COMPILER_DIR}/code.s" ".${FUNCTION}" > "${INPUT}"; rm ${COMPILER_DIR}/code.s; powerpc-linux-gnu-as "${INPUT}" -o "${OUTPUT}"'
```
ethteck, line 282: these should be able to be piped, `;`d, and/or `&&`d together and put all into the compiler flags

backend/coreapp/compilers.py
ethteck: I just realized this flag (and the special class) can disappear anyway once the other fields disappear

backend/coreapp/compilers.py
```diff
@@ -747,6 +791,8 @@ def preset_from_name(name: str) -> Optional[Preset]:
         MWCC_30_137,
         "-O4,p -enum int -lang c99 -Cpp_exceptions off -gccext,on -gccinc -interworking -gccdep -MD",
     ),
+    # MACOS9
+    Preset("The Sims 1", MWCPPC_24, "-lang=c++ -O2 -str pool -g"),
```
ethteck, line 795: I think The Sims is probably better, since that's the official name. Also, I think you were saying earlier it's -O3?

ethteck (changes requested):
Exciting to add a new platform and new compilers to the site! As I've mentioned before, I think this can be cleaned up a bit to remove duplicate code and avoid any new hacks. Otherwise, looks great!

ethteck (approved):

mkst (approved):

PR #456, opened 2022-05-31
Fix rodata references not showing up on RHS

simonlindholm:
It was hitting https://code.djangoproject.com/ticket/27813. Fixed by casting memoryview to bytes, just like at the other place `elf_object` is accessed.

(This fix is completely untested but I'm confident.)


ethteck:
thoynksh

PR #462, opened 2022-06-03
[WIP] Add Mac OS X platform and compilers

ChrisNonyminus:

PR #463, opened 2022-06-03
use grid instead of flex in PlatformSelect

nanaian:
Updates PlatformSelect to use a grid of three columns. This should prevent [whatever happened in #462](https://github.com/decompme/decomp.me/pull/462#issuecomment-1145940121), and also prevents the cells from growing to more than one column when there are less than three on a row (see: last cell width in image below).

image


ethteck:
thoynskh :)

PR #467, opened 2022-06-06
Add preset for Mario Kart Double Dash!!

SwareJonge:
Haven't tested this fully but so far these settings have worked for the game and JSystem sections

### DIFF
--- a/backend/coreapp/compilers.py
+++ b/backend/coreapp/compilers.py
@@ -737,6 +737,11 @@ _all_presets = [
         "-lang=c++ -g -Cpp_exceptions off -RTTI off -fp hard -fp_contract on -O4,p -maxerrors 1 -str reuse,pool,readonly -char unsigned -enum int -use_lmw_stmw on -inline off",
     ),
     Preset(
+        "Mario Kart Double Dash!!",
+        MWCC_247_105,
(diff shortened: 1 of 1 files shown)


backend/coreapp/compilers.py
```diff
@@ -736,6 +736,11 @@ def preset_from_name(name: str) -> Optional[Preset]:
         MWCC_247_92,
         "-lang=c++ -g -Cpp_exceptions off -RTTI off -fp hard -fp_contract on -O4,p -maxerrors 1 -str reuse,pool,readonly -char unsigned -enum int -use_lmw_stmw on -inline off",
     ),
+    Preset(
+        "Mario Kart Double Dash!!",
```
ethteck, line 740: It looks like the official non-stylized name is `Mario Kart: Double Dash` with the colon and without the exclamation points. I think we should use that one instead. Otherwise, looks good!

SwareJonge: I believe it's Mario Kart: Double Dash!! but honestly i don't mind the name :P

ethteck: the stylized name includes the semicolons, but the non-stylized name doesn't https://www.wikidata.org/wiki/Q5721431
I just think the exclamation points look kinda loud/out of place on the site

PR #472, opened 2022-06-09
Highlight tokens in the diff view on click

simonlindholm:
Poor man's register highlighting: run a regex client side to tokenize lines and make it possible to click tokens to highlight them on that side of the diff. (You can highlight different tokens on both sides, in order to trace register diffs.) Double-clicking highlights a token on both sides. Only one token can be highlighted at a time, and clicking a non-token clears all highlighting.

(The double-click behavior is kinda undiscoverable and unideal... I could imagine selecting the corresponding token on the other side automatically when the lines match up completely, or showing a help message about the double click functionality when the user highlights the same thing on both sides manually, or moving the highlight-only-on-one-side functionality to under a modifier key.)

### DIFF
--- a/frontend/src/components/Diff/Diff.module.scss
+++ b/frontend/src/components/Diff/Diff.module.scss
@@ -6,11 +6,6 @@
     flex-direction: column;
 }
 
-.highlighted {
-    background-color: #777 !important;
-    box-shadow: 0 0 2px #fff;
-}
-
 .bodyContainer {
     flex-grow: 1;
 }
@@ -133,3 +128,12 @@
 .rotation6 { color: lightcyan; }
 .rotation7 { color: lightgreen; }
 .rotation8 { color: grey; }
+
+.highlighted {
+    color: white;
+    background-color: #aa8b00;
+}
+
+.highlightable {
+    cursor: pointer;
+}
--- a/frontend/src/components/Diff/Diff.module.scss
+++ b/frontend/src/components/Diff/Diff.module.scss
@@ -72,6 +72,8 @@
 
         padding: 0 1em;
 
+        cursor: default;
+
         &.highlight {
             background-color: rgba(255, 255, 255, 0.05);
 
--- a/frontend/src/components/Diff/Diff.module.scss
+++ b/frontend/src/components/Diff/Diff.module.scss
@@ -6,6 +6,11 @@
     flex-direction: column;
 }
 
+.highlighted {
+    background-color: #777 !important;
+    box-shadow: 0 0 2px #fff;
+}
+
 .bodyContainer {
     flex-grow: 1;
 }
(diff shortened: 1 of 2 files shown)


frontend/src/components/Diff/Diff.tsx
```diff
@@ -89,7 +139,15 @@ const innerElementType = forwardRef<HTMLUListElement, HTMLAttributes<HTMLUListEl
 innerElementType.displayName = "innerElementType"
 
 function DiffBody({ diff, fontSize }: { diff: api.DiffOutput, fontSize: number | undefined }) {
-    return <div className={styles.bodyContainer}>
+    const setHighlightAll = value => {
+        highlighter1.setValue(value)
+        highlighter2.setValue(value)
+        highlighter3.setValue(value)
+    }
```
nanaian: Can you move this function to below the `highlighterX` declarations?

nanaian (approved):
Good rich diff feature. I can't say I love the implementation details-- there's a lot going on, and it would be better to actually parse the asm lines, but it's definitely good enough for now :shipit: 

We should definitely consider refactoring the DiffCell/FormatText etc to their own files if this is to get more complex in the future

frontend/src/components/Diff/Diff.tsx
```diff
@@ -52,23 +52,24 @@ function FormatDiffText({ texts, highlighter }: {
     highlighter: Highlighter
 }) {
     let index = 0
-    return <> {
+    return <>{
```
simonlindholm, line 55: this space actually appears in the output, and I think we want it there. Or some replacement CSS

frontend/src/components/Diff/Diff.tsx
```diff
@@ -52,23 +52,24 @@ function FormatDiffText({ texts, highlighter }: {
...
         texts.map(t =>
             Array.from(t.text.matchAll(RE_TOKEN)).map(match => {
-                const text = match[0], isToken = !match[1]
-                const key = index++
-                let className = undefined
+                const text = match[0]
+                const isToken = !match[1]
+                const key = index++ // FIXME: using index as key is an antipattern
```
simonlindholm, line 60: (I disagree in this case since the asm doesn't change dynamically, and there isn't any other key to use.)

frontend/src/components/Diff/Diff.module.scss
```diff
@@ -133,3 +128,12 @@
...
 .rotation7 { color: lightgreen; }
 .rotation8 { color: grey; }
+
+.highlighted {
+    color: white;
+    background-color: #aa8b00;
+}
+
+.highlightable {
+    cursor: pointer;
```
simonlindholm, line 140: I also experimented with this and felt it a bit distracting. Though it's true that it does add discoverability

nanaian: I definitely think the discoverability is worth it

ethteck (approved):
lgtm! thanks for this. It works well for such a simple addition. I'll leave it to you two to decide when to merge

snuffysasa:
excited to try this

PR #477, opened 2022-06-12
disallow_untyped_defs for mypy

ethteck:
Added `disallow_untyped_defs = True` to our mypy rules.

I'd appreciate it if a couple people go over this. Particularly any time I added `Any` or left a `# TODO` I'm less sure about.


backend/coreapp/serializers.py
```diff
@@ -55,18 +57,18 @@ def serialize_profile(request: Request, profile: Profile, small=False):
...
 if TYPE_CHECKING:
-    ProfileFieldBaseClass = serializers.RelatedField[Profile, str, str]
+    ProfileFieldBaseClass = serializers.RelatedField[Profile, str, dict[str, Any]]
 else:
     ProfileFieldBaseClass = serializers.RelatedField
 
 
 class ProfileField(ProfileFieldBaseClass):
-    def to_representation(self, profile: Profile):
+    def to_representation(self, profile: Profile) -> dict[str, Any]:
```
nanaian, line 66: Should this be `Dict`?

backend/coreapp/serializers.py
```diff
@@ -146,23 +150,38 @@ class Meta:
             "platform",
         ]
 
-    def get_project(self, scratch: Scratch):
+    # TODO: Unused
+    def get_project(self, scratch: Scratch) -> Optional[str]:
```
ethteck, line 154: @nanaian do you know why these are here? I don't think we need them on the serializer, and they seem unused

backend/coreapp/serializers.py
```diff
@@ -146,23 +150,38 @@ class Meta:
...
+            return reverse(
+                "project-detail",
+                args=[scratch.project_function.project.slug],
+                request=self.context["request"],  # type: ignore
+            )
+        return None
 
-    def get_project_function(self, scratch: Scratch):
+    # TODO: Unused
+    def get_project_function(self, scratch: Scratch) -> Optional[str]:
```
ethteck, line 167: this one too

backend/coreapp/views/scratch.py
```diff
@@ -406,12 +406,15 @@ def claim(self, request, pk):
...
     @action(detail=True, methods=["POST"])
-    def fork(self, request, pk):
+    def fork(self, request: Request, pk: str) -> Response:
         parent: Scratch = self.get_object()
 
-        request_data = (
-            request.data.dict() if isinstance(request.data, QueryDict) else request.data
-        )
+        # TODO Needed for test_fork_scratch test?
+        if isinstance(request.data, QueryDict):  # type: ignore
```
ethteck, line 413: this whole thing seems wrong too

nanaian (approved):

backend/coreapp/serializers.py
nanaian: these are not unused
https://www.django-rest-framework.org/api-guide/fields/#serializermethodfield

nanaian (changes requested):
Remove those `// TODO: Unused` comments; they're not unused

backend/coreapp/views/scratch.py
nanaian: IIRC when testing, `request.data` was a QueryDict, not a real dict, even though the stubs said it was supposed to be a dict, and the `**request_data` on L419 didn't work with the QueryDict

backend/coreapp/serializers.py
ethteck: okay, thanks. They used to not always return a value (the function would fall through to nothingness) - is the `return None` I added at the end of them (and the return type `str`) correct?

nanaian: yep LGTM

backend/coreapp/views/scratch.py
ethteck: yeah I saw the same thing. I'm just confused why that's the case..

nanaian: [it appears that DRF varies the type based on what type of data is POSTed, which is gross](https://stackoverflow.com/a/70230790)

nanaian (approved):

backend/coreapp/tests.py
```diff
@@ -33,7 +33,7 @@
 from .models.scratch import CompilerConfig, Scratch
 
 
-def requiresCompiler(*compilers: Compiler):
+def requiresCompiler(*compilers: Compiler) -> Callable[..., Any]:
```
simonlindholm, line 36: not that it matters whatsoever in a test, but you could do `Callable[[F], F]` where `F = TypeVar('F', bound=Callable[..., Any])` for more accurate typing. (The `bound=` part not too important.)

simonlindholm (approved):
LGTM. The remaining `Any`s look annoying to get rid of, but now they're at least explicit!

PR #488, opened 2022-07-14
MWCC line number fix

SeekyCt:
In MWCC versions 3.0+, using `#line` directives to override line numbers in the dwarf output won't work unless the file given in the directive exists (the contents of it don't matter). This PR changes the second directive from `src.c` to `code.c` since that file already exists, and creates an empty `ctx.c` file to fix the first directive.

### DIFF
--- a/backend/coreapp/compiler_wrapper.py
+++ b/backend/coreapp/compiler_wrapper.py
@@ -140,8 +140,7 @@ class CompilerWrapper:
             # Fix for MWCC line numbers in GC 3.0+
             if compiler.is_mwcc:
                 ctx_path = sandbox.path / "ctx.c"
-                with ctx_path.open("w") as f:
-                    pass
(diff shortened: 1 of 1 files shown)


backend/coreapp/compiler_wrapper.py
```diff
@@ -131,12 +131,18 @@ def compile_code(
...
+                f.write('#line 1 "code.c"\n')
                 f.write(code)
                 f.write("\n")
 
             cc_cmd = compiler.cc
 
+            # Fix for MWCC line numbers in GC 3.0+
+            if compiler.is_mwcc:
+                ctx_path = sandbox.path / "ctx.c"
+                with ctx_path.open("w") as f:
```
mkst, line 143: I think this can be `ctx_path.touch()` instead?

nanaian (approved):

PR #490, opened 2022-07-19
Add Nintendo 3DS platform

EstexNT:


simonlindholm:
We've had bad experiences with wine in decomp.me wrt sandboxing, startup time, and breakage on updates, with a recent move towards replacing wine by https://github.com/Treeki/WiBo. Hacking that to support armcc would be doable but require some effort. Are there Linux versions of the compiler available? I'm seeing some at https://developer.arm.com/downloads/-/legacy-compilers, though no 4.0 and only a subset of the 4.1 build numbers (plus some build numbers that aren't in the zip).

SinaMegapolis:
Nintendo made a deal with ARMCC developers so that specific ARMCC 4.1 builds (from 4.1 b9xx to 5.2?) were only given to Nintendo 3DS developers.
To my knowledge both the public and the proprietary 4.1 builds only work on Windows, specifically best supported for Windows XP, Vista, and 7. They don't come with Linux versions.

SinaMegapolis:
I actually think the reason ARMCC for 3DS doesn't have linux versions is that it's meant to work with Visual Studio, as it lets the developers have more flexibility when it comes to developing CROs and CRO based build systems

ethteck:
Merging for now (using wine) and will hopefully migrate it to wibo soon

PR #484, opened 2022-07-03
Replace wine with wibo

ethteck:

PR #495, opened 2022-07-29
Revert "Revert "Replace wine with wibo (#484)""

ethteck:
We'll merge this once we are more sure we support all the psx / GC / arm compilers.


ethteck:
Closing - will be handled in #503

PR #496, opened 2022-07-29
Remove 'psyq gcc w/ gnu as' compilers (repoint scratches to psyq asse…

mkst:
…mbler version)

I'm unable to test locally :(

### DIFF
--- a/backend/coreapp/migrations/0024_remove_psyq_as_compiler.py
+++ b/backend/coreapp/migrations/0024_remove_psyq_as_compiler.py
@@ -1,0 +1,36 @@
+import django.db.migrations.operations.special
+from django.apps.registry import Apps
+from django.db import migrations
+from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+
+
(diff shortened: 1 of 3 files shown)


backend/coreapp/migrations/0024_remove_psyq_as_compiler.py
```diff
@@ -0,0 +1,36 @@
...
+    """
+
+    compiler_map = {
+        "gcc2.7.2-psyq": "psyq4.1",  # conflating as both 4.0 and 4.1 are gcc2.7.2
+        "gcc2.8.1-psyq": "psyq4.3",
+        "gcc2.95.2-psyq": "psyq4.6",
+    }
+
+    Scratch = apps.get_model("coreapp", "Scratch")
+    for row in Scratch.objects.all():
```
ethteck, line 19: We can't do all() on Scratch anymore without causing memory issues. You'll have to run a filter instead (filter based on compiler in a loop of the keys in your map?) - see the previous manual migration for an example

PR #505, opened 2022-08-10
Fix incorrect line numbers on NDS MWCC

shinyquagsire23:
NDS MWCC outputs correct line numbers to stdout in the following format
```
Z:\var\folders\mv\5b1057tx3pgbq60mm6p07y0h0000gn\T\tmpreq_0h7l\ctx.c:7: note:     (corresponding #line reference)
```

Usually this gets interleaved with stderr in the command line, 
```
Z:\var\folders\mv\5b1057tx3pgbq60mm6p07y0h0000gn\T\tmpreq_0h7l\code.c:13: warning: illegal implicit conversion from 'int' to
Z:\var\folders\mv\5b1057tx3pgbq60mm6p07y0h0000gn\T\tmpreq_0h7l\code.c:13: warning: 'struct asdf_struct *'
Z:\var\folders\mv\5b1057tx3pgbq60mm6p07y0h0000gn\T\tmpreq_0h7l\ctx.c:7: note:     (corresponding #line reference)
```

...but split stdout/stderr makes it unusable since all the notes get tacked onto the end.

This PR tries to take each `corresponding #line reference` and match it up with (usually multiple) stderr lines, so that the output is as expected:
```
ctx.c:7: warning: illegal implicit conversion from 'int' to
ctx.c:7: warning: 'struct asdf_struct *'
```

### DIFF
--- a/backend/coreapp/compiler_wrapper.py
+++ b/backend/coreapp/compiler_wrapper.py
@@ -179,7 +179,7 @@ class CompilerWrapper:
                 if e.stdout:
                     msg = f"{e.stdout}\n{e.stderr}"
                 else:
-                    msg = e.stderr
+                    msg = f"{e.stderr}"
 
                 logging.debug("Compilation failed: %s", msg)
                 raise CompilationError(CompilerWrapper.filter_compile_errors(msg))
@@ -192,7 +192,7 @@ class CompilerWrapper:
                 if compile_proc.stdout:
                     msg = f"{compile_proc.stdout}\n{compile_proc.stderr}"
                 else:
-                    msg = compile_proc.stderr
+                    msg = f"{compile_proc.stderr}"
                 error_msg = "Compiler did not create an object file: %s" % msg
                 logging.debug(error_msg)
                 raise CompilationError(error_msg)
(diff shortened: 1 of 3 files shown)


ethteck:
Could you run mypy and black to fix the issues they're reporting?

ethteck:
Also, for these compilers, couldn't we just not split stderr/stdout? I'm not really a fan of complicated string-parsing functions like this

ethteck:
additionally, maybe the compiler has flags that influence how errors/warnings are reported. that would also be really useful to look into

shinyquagsire23:
Yeah I did a check for compiler flags (nothing in `--help` but I also checked in IDA) before I implemented this, no dice. Though I might check in IDA again just to be sure because I can't say I like the idea of having to use regex for cleaner output.

shinyquagsire23:
Oh great I found a command line option, `-pragma msg_show_realref`. I'll refactor to that then.

backend/coreapp/compiler_wrapper.py
```diff
@@ -179,7 +179,7 @@ def compile_code(
                 if e.stdout:
                     msg = f"{e.stdout}\n{e.stderr}"
                 else:
-                    msg = e.stderr
+                    msg = f"{e.stderr}"
```
ethteck, line 182: This format string isn't needed if we're just setting msg to one variable. But also I think we can change this whole section to just `msg = e.stdout` since stderr now no longer exists.

backend/coreapp/compiler_wrapper.py
```diff
@@ -192,7 +192,7 @@ def compile_code(
                 if compile_proc.stdout:
```
ethteck, line 192: here too - this should just be `msg = compile_proc.stdout`

ethteck (changes requested):

PR #511, opened 2022-08-19
Add SN64 compiler binaries

marijnvdwerf:


ethteck:
It'd be nice if you could produce a tar with correct files / permissions / etc that we can just extract, in the future. but lgtm for now. merging

Issue #389, opened 2022-02-21
Cute names for anonymous profiles


ethteck:

nanaian:
and avatars! like google docs

nanaian:
porting https://www.npmjs.com/package/make-it-name might be cute

ethteck:
thanks @ConorBobbleHat :)

PR #513, opened 2022-08-30
Nicknames for anonymous users

ConorBobbleHat:
As per #389.

I figured the best way to go about this would be pretty subjective; hence drafting early.

Current changes:
- [x] New anonymous users are assigned a pseudonym, which is used as a subsitute for a username whenever appropriate. Right now, the logic for this is more or less a carbon copy of https://www.npmjs.com/package/make-it-name (as suggested in the original issue) - I'm more than happy to replace / complement it with another scheme if needed.

TODO:
- [ ] Profile pictures for anonymous users. Not sure what the best way to go about this is. A rainbow of random-colored frogs? Take inspiration from Google's anonymous animals? Procedural abstract art? (A mix of all the above? :))


ethteck:
LGTM! as discussed in discord, the only thing remaining on my end is to give the new names to any existing anon profiles that use the site after the change is made. I think it'd make sense to check this on scratch creation / save endpoints, if not any time they access the site. 

As for random image, I know less about the options for doing that and what's simple / viable. My initial thought is to colorize/modify a base image (the frog?), using a hash of the anon name as the seed for the modifications. that way the client just has to download one image and it gets warped client-side. but in general, I think this part can be done/expanded upon in a future PR

backend/coreapp/models/profile.py
```diff
@@ -1,9 +1,26 @@
...
+from typing import Optional, Iterable
+import os
+import json
+import random
 
 from django.contrib.auth.models import User
 from django.db import models
 from django.utils import timezone
 
+with open(os.path.join(os.path.dirname(__file__), "pseudonym_data.json")) as f:
```
mkst, line 10: Maybe prefer Path from pathlib?

backend/coreapp/serializers.py
```diff
@@ -26,6 +26,7 @@ def serialize_profile(
             "is_anonymous": True,
             "id": profile.id,
             "is_online": profile.is_online(),
+            "username": f"{profile.pseudonym} (anon)" if profile.pseudonym else None,
```
mkst, line 29: It's early and my brain is still warming up, but would this return pseudonyms for registered used instead of their GitHub username?

frontend/src/components/user/UserLink.tsx
```diff
@@ -21,9 +21,10 @@ export default function UserLink({ user }: Props) {
     const userIsYou = api.useUserIsYou()
 
     if (api.isAnonUser(user)) {
+        const username = api.userHasName(user) ? user.username : "anon"
         return <a className={styles.user}>
             <UserAvatar user={user} />
-            <span>{userIsYou(user) ? "anon (you)" : "anon"}</span>
+            <span>{userIsYou(user) ? `${username} (you)` : username}</span>
```
nanaian, line 27: Happy to drop "(you)" entirely since we display the pseudonym in the nav now

backend/coreapp/serializers.py
nanaian: This is in a 'if profile.user is None' block, so no

frontend/src/lib/api.ts
```diff
@@ -153,6 +153,7 @@ export interface AnonymousUser {
     is_anonymous: true
     id: number
     is_online: boolean
+    username: string
```
nanaian, line 156: This can be null, so we should type it that way.

However, I think reducing the number of states username can be in would be good. Could we make it so username is "anon" (or _always_ populated with a generated name) such that we can get rid of userHasName checks?

nanaian (dismissed):

nanaian:
RE: avatars, a client-side coloured frog should suffice. We can do this using the existing frog SVG and CSS variables to colour it.

frontend/src/lib/api.ts
ethteck: I actually agree. The issue is that I don't want to burn a bunch of potentially cool names on the thousands(?) of never-to-be-used-again anonymous profiles. But we can set all existing profiles to a default name ("anon") and then, when the site is used by an anon profile, if the profile's username is the default one, we generate one instead?

ethteck: after discussion in discord, this will always be non-null now. userHasName can go away still

backend/coreapp/models/profile.py
ConorBobbleHat: Implemented.

frontend/src/components/user/UserLink.tsx
ConorBobbleHat: Dropped.

frontend/src/lib/api.ts
ConorBobbleHat: Gone away.

ConorBobbleHat:
As per discord, colored frogs will be in another PR.

ethteck (approved):
Looks great! I'm super excited for this change. Thank you for your contribution

backend/coreapp/models/profile.py
```diff
@@ -1,9 +1,25 @@
...
+from typing import Optional, Iterable
+from pathlib import Path
+import json
+import random
 
 from django.contrib.auth.models import User
 from django.db import models
 from django.utils import timezone
 
+with open(Path(__file__).resolve().parent / "pseudonym_data.json") as f:
```
nanaian, line 10: you can do .open here

backend/coreapp/serializers.py
nanaian, line 29: psuedonym cannot be null so this check is useless

ConorBobbleHat:
no problem!

nanaian (approved):

nanaian (changes requested):
a

PR #522, opened 2022-09-12
Gizzard Shad

nanaian:
image


ethteck:
Lol but the whole fish is gizzard shad. Gizzard isn't an adjective describing the shad! Hahahahaha

Can we also add Grey Sole and Yellow Perch? So I guess just sole and perch as animals if those don't exist, and grey and yellow as adjectives

PR #541, opened 2022-09-18
#539 again

nanaian:
swc strikes back


frontend/src/lib/api.ts
```diff
@@ -488,7 +488,11 @@ export function useCompilation(scratch: Scratch | null, autoRecompile = true, au
             setCompileRequestPromise(null)
             setIsCompilationOld(false)
         }).catch(error => {
-            setCompilation({ "errors": error.json?.detail, "diff_output": null })
+            if (error instanceof ResponseError && error.code == "Compiler") {
```
ethteck, line 491: We can return the following error codes from a compile endpoint request:
`Compiler`, `nm`, `Sandbox`, `objdump`, `Diff` - so we'll want to handle any of those. If you prefer a different error format that requires less frontend checks, I'm happy to re-think that

ethteck (changes requested):

PR #542, opened 2022-09-18
Projects list, creation, settings pages

### LINKED ISSUE #533: PR creation updates
e.g. https://github.com/pmret/papermario/pull/774

- Delete s files
- Authors are not properly given attribution
- Link to scratch in PR body

nanaian:
- [x] Projects list
- [x] Project creation page
  - [x] Upload project icon (need MEDIA_ROOT set in prod!)
- [x] Project settings page
    - [x] Change description
    - [x] Add/remove members


backend/coreapp/models/project.py
```diff
@@ -41,7 +43,10 @@ def import_functions(self) -> None:
             import_config.execute_import()
 
     def is_member(self, profile: Profile) -> bool:
-        return ProjectMember.objects.filter(project=self, profile=profile).exists()
+        for member in self.members():
```
ethteck, line 46: This should stay as a filter imo - it's much more Django-ish

backend/coreapp/serializers.py
```diff
@@ -212,24 +215,52 @@ class Meta:
...
+        return instance
+
     def get_members(self, project: Project) -> List[Dict[str, Any]]:
         return [
-            serialize_profile(self.context["request"], member.profile, True)
+            serialize_profile(self.context["request"], member.user.profile, True)
             for member in project.members()
         ]
 
+    def get_most_common_platform(self, project: Project) -> Optional[str]:
```
ethteck, line 250: Can you explain the motivation for this? Why would a project even have more than one platform? I think if you're decompiling a game that shares platforms, those should be considered multiple projects on the site. Maybe there's a scenario you had in mind I'm not seeing though

backend/coreapp/views/project.py
```diff
@@ -239,6 +302,50 @@ def pr(self, request: Request, pk: str) -> Response:
         )
         return Response({"url": response.html_url})
 
+    @action(detail=True, methods=["GET", "PUT"])
+    def members(self, request: Request, pk: str) -> Response:
```
ethteck, line 306: So the client has to send a full list of project members each time a single person is to be added? I'm not sure I like that model. Can't we just have one command to add members and another to remove them? or even keep the single PUT but have two parameters, one for people to add and one for people to remove?

backend/mypy.ini
```diff
@@ -29,5 +28,35 @@ plugins =
     mypy_django_plugin.main,
     mypy_drf_plugin.main
 
+mypy_path = $MYPY_CONFIG_FILE_DIR
+
 [mypy.plugins.django-stubs]
 django_settings_module = decompme.settings
+
+[mypy-m2c.*]
```
ethteck, line 36: We shouldn't need settings blocks for either of these since we're not even scanning for them, right? (we shouldn't be running mypy on either of them if we are)

backend/pyproject.toml
```diff
@@ -26,6 +26,10 @@ PyGithub = "^1.55"
 drf-extensions = "^0.7.1"
 tqdm = "^4.62.3"
 tzdata = "^2022.1"
+Pillow = "^9.2.0"
```
ethteck, line 29: What's Pillow for? or html-json-forms?

ethteck (changes requested):
Some comments just reading the code

backend/coreapp/models/project.py
nanaian: IIRC this broke tests, but I can revisit

backend/coreapp/serializers.py
nanaian: It's basically just to show the game's platform on the frontend, "most common" here almost certainly always means _the_ platform, it's just describing what the method actually does. Like Github does with repo languages

backend/coreapp/views/project.py
nanaian: Bear in mind this "full list" is just usernames. 

> Can't we just have one command to add members and another to remove them?

We could implement this as a ViewSet, I guess?
`POST /members` - add member
`DELETE /members/username` - remove member
`PATCH /members/username` - change member's role (later)

> or even keep the single PUT but have two parameters, one for people to add and one for people to remove?

Violates PUT semantics

backend/coreapp/serializers.py
ethteck: okay. I think it's needlessly complex though. A project can just have one platform, imo. doing this whole calculation is unnecsesary. We can just grab the first object and use its platform

backend/mypy.ini
nanaian: dmypy doesn't support `follow-imports=false` so we do scan them. see the commit which originally added this

backend/pyproject.toml
nanaian: django image field

backend/coreapp/models/project.py
nanaian: Breaks tests so leaving as-is (profile.user is mocked and explodes)

backend/coreapp/serializers.py
nanaian: > We can just grab the first object and use its platform

done :+1:

ethteck (approved):

PR #568, opened 2022-10-22
Fixed the link of LICENSE.

Akash190104:
Added the link to the MIT License for the ease of people to read. @EllipticEllipsis Please review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,4 +15,4 @@ To learn more, see [CONTRIBUTING.md](docs/CONTRIBUTING.md)
 
 
 ## License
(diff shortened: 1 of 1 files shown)


ethteck:
Thanks for your time and effort, but I don't really think this is necessary given our license file is in a standard place and format. Not to mention, the link you added is a web link to an absolute location, rather than a relative one like the other markdown links we use. If you are interested in contributing to the project, though, there are plenty of things we could use help with in our Issues section. Feel free to join the Discord as well, and we can make some suggestions if you want.

Akash190104:
@ethteck Thanks for the detailed explanation. I would find other issues to work on.

Issue #439, opened 2022-04-13
Use CompilerConfig model in Scratch model


bates64:
A little refactor.
Make sure to not break the existing create scratch endpoint's interface (#148 would help here!)

nanaian:
@ethteck did we get anywhere with this

ethteck:
I've been working on it

ethteck:
This isn't needed anymore since we store Presets in the backend, which are essentially CompilerConfig

PR #570, opened 2022-10-26
Pascal language support (ido7.1Pascal compiler)

EllipticEllipsis:
Add Pascal support to IDO 7.1. Includes
- New recomp version including the 7.1 Pascal frontend `upas`
- New compiler options: minimal-flags C IDO for both 5.3 and 7.1, and Pascal for 7.1 (all expected to be useful for IDO decomp)
- New presets for 3 sorts of known IDO code.
- Automatically change the `code.c` extension to `p` when necessary.
- Remove alphabetical ordering from the compilers, since an ordering more based on frequency of use seems like a better idea.


backend/coreapp/compilers.py
```diff
@@ -123,10 +124,7 @@ def from_id(compiler_id: str) -> Compiler:
 
 @cache
 def available_compilers() -> List[Compiler]:
-    return sorted(
```
ethteck, line 126: I think this was sorted intentionally. Is _compilers.values a static order?

EllipticEllipsis: `_compilers` is an `OrderedDict`, so I assume it's just ordered the way the definition is, that at least reflects what I see locally.

nanaian (changes requested):
There's frontend work to be done here, namely:
- Expose the compiler `language` someplace, perhaps grouping the compiler picker by language.
- Don't try to syntax-highlight Pascal as C++
  - Happy to merge this with zero Pascal syntax highlighting (i.e. remove lang-cpp), although of course it would be better to actually highlight Pascal (we could do this in the future though)
  - Perhaps use a different lang package for C vs C++

backend/coreapp/compiler_wrapper.py
```diff
@@ -133,14 +133,18 @@ def compile_code(
...
         with Sandbox() as sandbox:
-            code_path = sandbox.path / "code.c"
+            code_file = "code.c"
+            if compiler.language == "Pascal":
+                code_file = "code.p"
+
+            code_path = sandbox.path / code_file
             object_path = sandbox.path / "object.o"
             with code_path.open("w") as f:
                 f.write('#line 1 "ctx.c"\n')
```
nanaian, line 143: This needs to use the extension also

nanaian (changes requested):

backend/coreapp/compilers.py
```diff
@@ -48,6 +48,7 @@ class Compiler:
     is_ido: ClassVar[bool] = False
     is_mwcc: ClassVar[bool] = False
     needs_wine = False
+    language: str = "C"
```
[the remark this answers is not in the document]
EllipticEllipsis: It doesn't seem to like that, complains that "C" is not defined. Tried both `|` and `Union` syntax.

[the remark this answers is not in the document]
EllipticEllipsis: Got it to work as an enum.

backend/coreapp/compiler_wrapper.py
```diff
@@ -133,11 +133,10 @@ def compile_code(
...
         with Sandbox() as sandbox:
-            code_file = "code.c"
-            ctx_file = "ctx.c"
-            if compiler.language == "Pascal":
-                code_file = "code.p"
-                ctx_file = "ctx.p"
+            ext = compiler.language.get_file_extension()
+            code_file = f"code.{ext}"
+            ctx_file = f"ctx.{ext}"
+            print(code_file)
```
nanaian, line 139: Remove this line

backend/coreapp/views/scratch.py
```diff
@@ -463,9 +464,12 @@ def export(self, request: Request, pk: str) -> HttpResponse:
             zip_f.writestr("metadata.json", json.dumps(metadata, indent=4))
             zip_f.writestr("target.s", scratch.target_assembly.source_asm.data)
             zip_f.writestr("target.o", scratch.target_assembly.elf_object)
-            zip_f.writestr("code.c", scratch.source_code)
+
+            language = Language.PASCAL if scratch.compiler == "ido7.1pascal" else "C"
```
nanaian, line 468: Can you chuck a `# TODO refactor #439` on this line as its clearly icky

nanaian (changes requested):

backend/coreapp/compilers.py
```diff
@@ -296,6 +308,15 @@ def preset_from_name(name: str) -> Optional[Preset]:
     cc='IDO_CC="${COMPILER_DIR}/cc" "${COMPILER_DIR}/cc" -c -Xcpluscomm -G0 -non_shared -Wab,-r4300_mul -woff 649,838,712 -32 ${COMPILER_FLAGS} -o "${OUTPUT}" "${INPUT}"',
 )
 
+# Pascal IDO
+IDO71PASCAL = IDOCompiler(
+    id="ido7.1Pascal",
```
simonlindholm, line 313: I'm confused. This line says "ido7.1Pascal",

backend/coreapp/views/scratch.py
```diff
@@ -463,9 +464,13 @@ def export(self, request: Request, pk: str) -> HttpResponse:
             zip_f.writestr("metadata.json", json.dumps(metadata, indent=4))
             zip_f.writestr("target.s", scratch.target_assembly.source_asm.data)
             zip_f.writestr("target.o", scratch.target_assembly.elf_object)
-            zip_f.writestr("code.c", scratch.source_code)
+
+            # TODO refactor #439
+            language = Language.PASCAL if scratch.compiler == "ido7.1pascal" else "C"
```
simonlindholm, line 469: while this says "ido7.1pascal" with different capitalization. Is this a bug?

Also, this should really be using `coreapp.compilers.from_id(scratch.compiler).language`.

ethteck: Good catch

PR #580, opened 2022-11-03
(hopefully) fix macos9 compiler

ChrisNonyminus:
This should probably fix the issues that were going on with getting it to work.
What I had to do was update the assembler syntax converter, as its detection of branches wasn't working right (the regex was expecting 2 tabs when there was only one).

### DIFF
--- a/frontend/next.config.js
+++ b/frontend/next.config.js
@@ -29,7 +29,7 @@ const removeImports = require("next-remove-imports")({
 })
 const nextTranslate = require("next-translate")
 
-const mediaUrl = new URL(process.env.MEDIA_URL != null ? process.env.MEDIA_URL : "http://localhost")
+const mediaUrl = new URL(process.env.MEDIA_URL ?? "http://localhost")
 
 let app = withPlausibleProxy({
     customDomain: "https://stats.decomp.me",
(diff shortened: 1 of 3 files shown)


frontend/next.config.js
```diff
@@ -29,7 +29,7 @@ const removeImports = require("next-remove-imports")({
 })
 const nextTranslate = require("next-translate")
 
-const mediaUrl = new URL(process.env.MEDIA_URL ?? "http://localhost")
+const mediaUrl = new URL(process.env.MEDIA_URL != null ? process.env.MEDIA_URL : "http://localhost")
```
nanaian, line 32: Can you explain the reasoning for this change?

ChrisNonyminus: For some reason, the ``??`` won't mesh well with the latest javascript install on my machine.

nanaian: That's weird, what's your `node --version`?

ChrisNonyminus: v12.22.9

nanaian: Hm, `??` should be supported OK there. By "won't mesh well" what exactly do you mean? Did it complain of a syntax error?

ChrisNonyminus: Yes.

ChrisNonyminus: Updated node, removing the need for this workaround. Thus, I reverted the workaround.

nanaian (approved):

PR #585, opened 2022-11-11
Set the correct march for PS2 assembler

Mc-muffin:
PS2's  Emotion Engine has some custom instructions like `mult rd,rs,rt` and `madd` which weren't recognized as instructions with `-march=mips64`, so this PR just switches to `-march=r5900` which correctly handles the new instructions


mkst (approved):

ethteck:
We should probably handle the diff.py stuff here as well (diff_wrapper). see the convo here: https://discord.com/channels/897066363951128586/897077740476448768/1040547122757505074

PR #587, opened 2022-11-14
Some rodata-related functionality

EllipticEllipsis:
- Pull latest versions of asm-differ and m2c, asm-differ now allows for viewing rodata labels for PIC as well.
- Add flag to hide these if not desired to view them. I'm not very happy with how this is implemented at the moment: as @simonlindholm notes it's a bit of a flaw in the current setup that the flag name has to agree with the display name, among other issues.
- Massage the asm handling a bit to make it easier to create a scratch for a function with late_rodata in its asm.
- Add some more IDO presets, for 5.3.

### DIFF
--- a/backend/asm_differ/.gitrepo
+++ b/backend/asm_differ/.gitrepo
@@ -6,7 +6,7 @@
 [subrepo]
 	remote = https://github.com/simonlindholm/asm-differ
 	branch = main
-	commit = 01a895630c8ec585f56ce79b36e6bfd82ef55c9f
-	parent = f38775de5452dda85507901949ef31a39f09365b
+	commit = 37ecd6bd5bfc039008be57a78bb588b7fcfd5f1f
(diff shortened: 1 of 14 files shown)


ethteck:
this looks good. I feel like there's something else we're using as a way to control diff.py but I can't remember what it is atm :/ - we should move that to use this new DIFF_FLAG_PREFIX thing

EllipticEllipsis:
image
New commit does this, adds algorithm and moves the diff flags into their own thing (Yes, lot of duplication, but seems required unless we overhaul the `Flags` structs.)

simonlindholm:
Can we hide the subtitles? -DIFFno_show_rodata_refs is a purely internal identifier.

EllipticEllipsis:
image
Might need to tweak the style a bit if we do, looks like the boxes are too high without?

ethteck:
@nanaian thoughts on this? I'm happy to merge it as-is and we can improve the design later. whatever works

PR #588, opened 2022-11-15
Fix late_rodata edit

EllipticEllipsis:


simonlindholm:
Seems very unexpected for this function to modify the asm object passed by the caller. Can asm.data be stored in a local variable instead?

EllipticEllipsis:
I suppose that makes sense although it wastes a bit of time having to do the replacement every compile.

nanaian (approved):

PR #594, opened 2022-12-07
Add IRIX as a platform

EllipticEllipsis:
Also cleaned up the flags a bit. Possibly the Pascal one should be renamed to match, but I didn't want to fiddle with all the names again.

Going to try to stop agonising over the logo now!


EllipticEllipsis:
Oh, and big thanks to @pixel-stuck for the logo.

PR #603, opened 2022-12-14
A few issues

### LINKED ISSUE #249: Implement database schema for courses
(from my comment in #247)

### Course (has a 1-many relationship with CourseStep)
slug - the slug of the course
name - the name of the course
description - the description of the course

### CourseChapter
slug - the slug of the chapter
name  - the name of the chapter
index - the number of the chapter
course - the Course

### CourseScenario
slug - the slug of the scenario
name - the name of the step (could use scratch's name but feels better to store its own name in its object. plus this allows not worrying about the scratch changing and suddenly the name is different, etc)
index - which step in the course is this (will be used to gauge completion)
courseChapter - the CourseChapter it is part of
scratch - a Scratch which is the starting point for the step in terms of code / asm / context
content - some html or script or other kind of data which is the "course material" (probably best to start as just text for now)

### LINKED ISSUE #312: Plausible login event does not work
Probably because the window is closed immediately after it is triggered

### LINKED ISSUE #583: Show target asm even when scratch won't compile

ethteck:
Fixes #312 - removes login event from plausible
Fixes #249 - adds the course db models
Fixes #583 - allow viewing target asm even when the scratch doesn't initially compile

Update frontend & backend deps


backend/coreapp/models/course.py
```diff
@@ -0,0 +1,39 @@
...
+
+class CourseScenario(models.Model):
+    chapter = models.ForeignKey(
+        CourseChapter, on_delete=models.CASCADE, related_name="scenarios"
+    )
+    index = models.IntegerField()
+    slug = models.SlugField(max_length=100)
+    name = models.CharField(max_length=500)
+    content = models.TextField()
+    scratch = models.ForeignKey(Scratch, on_delete=models.CASCADE)
```
nanaian, line 36: Should probably be prevent deletion

backend/coreapp/views/scratch.py
```diff
@@ -381,7 +381,9 @@ def compile(self, request: Request, pk: str) -> Response:
         return Response(
             {
                 "diff_output": diff,
-                "errors": compilation.errors,
+                "compiler_output": compilation.errors,
+                "succeeded": compilation.elf_object is not None
```
nanaian, line 385: "success" probs better

Issue #593, opened 2022-12-06
Scratch ui takes a long time to load on scratches with very large contexts


ethteck:
Reported by @EpochFlame on Firefox, but I've noticed it too on Chrome as well.

I don't think this is purely a limitation of network speed, because this didn't used to manifest like this

example scratch: https://decomp.me/scratch/xSD1k

<img width="601" alt="image" src="https://user-images.githubusercontent.com/2985314/205992083-30dfd352-f502-45f9-9f9b-e21cdb749c61.png">

The request it's waiting on is the scratch api call, but there's only 640 kb of data to send and it takes a long time to recieve it.

ethteck:
During the waiting period, I have a window that looks like this (no loading bar, no indication of loading, a very long delay before anything appears on screen) 
<img width="965" alt="image" src="https://user-images.githubusercontent.com/2985314/205992956-2cd1c0d1-b31d-42c3-b950-99768837c37c.png">

EpochFlame:
worth noting it also does the above on iphone safari

nanaian:
Looks like this issue is in the ServiceWorker; I thought I removed it but I guess I failed
Likely cause of #595 too

Issue #545, opened 2022-09-18
CodeMirror useCompareExtension should be performed in a worker


nanaian:
See https://github.com/decompme/decomp.me/issues/535#issuecomment-1250334311

PR #605, opened 2022-12-15
Move logic of useCompareExtension to a Web Worker

ConorBobbleHat:
Implements #545.

### DIFF
--- a/frontend/.babelrc
+++ b/frontend/.babelrc
@@ -1,4 +1,4 @@
 {
     "presets": ["next/babel"],
-    "plugins": []
+    "plugins": ["@shopify/web-worker/babel"]
 }
--- a/frontend/src/lib/codemirror/useCompareExtension.ts
+++ b/frontend/src/lib/codemirror/useCompareExtension.ts
@@ -1,63 +1,20 @@
-import { RefObject, useEffect, useRef, useState } from "react"
+import { RefObject, useEffect, useState } from "react"
 
-import { EditorState, Compartment, Extension, Facet, Text } from "@codemirror/state"
-import { EditorView, gutter, GutterMarker } from "@codemirror/view"
-import { diff } from "fast-myers-diff"
+import { Compartment, Extension, Facet } from "@codemirror/state"
+import { EditorView, gutter, GutterMarker, ViewPlugin, ViewUpdate } from "@codemirror/view"
+import { createWorkerFactory } from "@shopify/web-worker"
 
 import styles from "./useCompareExtension.module.scss"
-
-function compareNullableText(a: Text | null, b: Text | null): boolean {
-    if (a === null || b === null) {
-        return a === b
-    } else {
-        return a.eq(b)
-    }
-}
 
 // State for target text to diff doc against
 const targetString = Facet.define<string, string | null>({
     combine: values => (values.length ? values[0] : null),
-})
-const targetText = Facet.define<Text, Text | null>({
-    combine: values => (values.length ? values[0] : null),
-    compare: compareNullableText,
-    compareInput: compareNullableText,
-})
-const targetTextComputer = targetText.compute([targetString], state => {
-    const s = state.facet(targetString)
-    if (typeof s === "string")
-        return Text.of(s.split("\n"))
-    return null
 })
 
 // Computed diff between doc and target
 type DiffLineMap = Record<number, GutterMarker>
 const diffLineMap = Facet.define<DiffLineMap, DiffLineMap>({
     combine: values => (values.length ? values[0] : {}),
-})
-const diffLineMapComputer = diffLineMap.compute(["doc", targetString], state => {
-    const s = state.facet(targetString)
-
-    if (typeof s !== "string")
-        return {}
-
-    const tokenizeSource = source => {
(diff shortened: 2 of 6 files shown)


ConorBobbleHat:
There is almost certainly a CodeMirror-ier way to accomplish some pieces of this :)

The only other thing worth noting is this PR makes use of @shopify/web-worker, which provides a neat RPC-style interface for web workers, and lets you just freely call and receive data from them without any more syntactic overhead than any other function call. That being said, it does require some changes to the build pipeline to function. If those don't float your boat, they aren't essential - we can switch to plain old web workers.

ConorBobbleHat:
oops, forget: also fixes #593

ethteck:
Very cool! Thanks for doing this :) I have no opinions on the build pipeline stuff - I guess one question is if it would interfere with swc minify if we choose to use that

frontend/src/lib/codemirror/useCompareExtension.ts
```diff
@@ -88,38 +45,55 @@ const diffGutter = gutter({
...
+                    map[i + 1] = marker
+                }
+            }
+
+            // Has our targetString been updated to a blank,
+            // and thus we should be showing no diff right now, while
+            // the view's been updating?
+            if (typeof this.view.state.facet(targetString) !== "string") {
+                map = {}
+            }
```
nanaian, line 90: When does this case occur? If the `targetString` is missing (null?) then wouldn't `diff` return `[]` and therefore `map` would be `{}`?

frontend/src/lib/codemirror/useCompareExtension.worker.ts
```diff
@@ -0,0 +1,16 @@
+import { diff } from "fast-myers-diff"
+
+declare type Vec4 = [number, number, number, number];
```
nanaian, line 3: `declare` here seems redundant

frontend/.babelrc
```diff
@@ -1,4 +1,4 @@
 {
     "presets": ["next/babel"],
-    "plugins": []
+    "plugins": ["@shopify/web-worker/babel"]
```
nanaian, line 3: You add `@shopify/web-worker/babel` as a plugin for all files. Will it only apply itself to workers? I feel like we should be selectively loading this plugin for `.worker.[jt]sx?` files only.

nanaian (commented):
This is really great, thank you! I have a few questions.

This web worker RPC Babel plugin is really awesome; will be bookmarking that for use elsewhere.

frontend/src/lib/codemirror/useCompareExtension.ts
```diff
@@ -88,38 +45,55 @@ const diffGutter = gutter({
...
-    ]
-}
+const createDiffWorker = createWorkerFactory(() => import("./useCompareExtension.worker"))
 
 // Extension that highlights lines in the doc that differ from `compareTo`.
 export default function useCompareExtension(viewRef: RefObject<EditorView>, compareTo: string): Extension {
     const [compartment] = useState(new Compartment())
+    const [diffLineMapCompartment] = useState(new Compartment())
+
+    const plugin = ViewPlugin.fromClass(class {
```
nanaian, line 60: Might be better to hoist the class to the top level and give it a name. This anonymous class confused me for a sec.

frontend/src/lib/codemirror/useCompareExtension.ts
ConorBobbleHat: tl;dr This is to handle the case where `compareTo` changes whilst `update` is executing.

During testing, it turned out that when the scratch is updated, the `update` function in useCompareExtension is fired before `isModified` is set to true (and thus `compareTo` is set to undefined). This means it would take two updates to get the diff to disappear instead of one.

I wasn't able to find a way to reorder the event handing to stop the diff calculation from occuring (and I'd love to remove this check if there is), but it turns out that await'ing the web worker reliably updates `compareTo` to its correct value - hence this check existing at the end of the function.

frontend/.babelrc
ConorBobbleHat: Based on [this](https://www.npmjs.com/package/@shopify/web-worker), it seems the Babel plugin is only used to collate and process calls to `createWorkerFactory`, and so needs to process all source files.

frontend/src/lib/codemirror/useCompareExtension.worker.ts
ConorBobbleHat: fair enough!

frontend/src/lib/codemirror/useCompareExtension.ts
ConorBobbleHat: will do

nanaian:
> I have no opinions on the build pipeline stuff - I guess one question is if it would interfere with swc minify if we choose to use that

@ethteck, yes, this will prevent us using swc to build in future. I doubt that time will come, to be honest, since we have other issues with swc such as it being unable to build codemirror properly. Therefore I'm happy to use Babel for syntax extensions like this!

nanaian:
Also, swc is extensible. So eventually as the ecosystem matures we might see a port of this Babel plugin

PR #614, opened 2022-12-19
Move CompilerConfig to Scratch

### LINKED ISSUE #439: Use CompilerConfig model in Scratch model

ethteck:
Closes #439


ethteck:
Closing this - the more I think about it, the more I prefer a flatter design for storing scratch metadata. But even if we do want to do this still, very little work was done here so far and the branch is severely out of date.

PR #618, opened 2022-12-19
Upgrade to Next.js 13

### LINKED ISSUE #610: Upgrade to Next.js 13
https://nextjs.org/docs/upgrading#upgrading-from-12-to-13

nanaian:
Resolves #610.


ethteck:
very coral; will test soon.

do you think we can benefit from migrating to the app/ stuff at some point?

nanaian:
The app directory is still in beta; it's not recommended we use it in production. Having said that, investigating if we can get much use out of server components would be a good plan in future

nanaian:
I don't think we should migrate our existing routing to app/ until its stable, but new features might want to use the app directory. It'll also reduce a lot of repeated code (for example, with the overall page layout)

PR #624, opened 2022-12-19
Use Poetry dependencies instead of subrepos

nanaian:


nanaian:
Just encountered an issue with asm-differ: `AttributeError: module 'Levenshtein' has no attribute 'opcodes'`
To fix this, create a new virtualenv with Poetry and reinstall everything: `poetry env remove --all && poetry install`

PR #625, opened 2022-12-19
Close problems panel (attempt 2)

### LINKED ISSUE #395: Hide Problems panel
The recently added panel for displaying compiler warnings and errors currently has no way of hiding. After taking a look at the warnings, you might want to ignore them, hiding the panel until "more issues" happen - which is not possible at the moment.

nanaian:
Closes #395


nanaian:
Implemented in #626 instead

PR #628, opened 2022-12-21
Implement timeouts for m2c and compiler invocations

ConorBobbleHat:
As this title suggests, this PR caps the runtime for all compiler / m2c calls. Currently, the timeout is five seconds for both.

There may well be a better place to put the compile timeout; as ever, happy to change :)


ethteck:
Thanks for this - it's great! Unfortunately it seems something with it breaks the windows build (but to be honest I couldn't care less about the windows build)

As for the other builds, the slowness of wine makes some build take longer than 5s. I think we could just increase the timeout when running tests, for now. but yeah this is revealing how crappy wine is for this project, which is a fun side-effect

nanaian:
Changes look good. The GitHub Actions runners aren't very fast, which means the tests fail. Could you make the timeout configurable via an environment variable (or a `.env` file)? We can default to a more liberal value like 15s so that the tests pass. We can then change the timeout at deploy-time based on how fast the server is.

ethteck:
Making them django settings that are in the .env makes sense to me. I think the default in the repo should be 10s for compilations and 5 for decompilations (if we want to separate the two). we still need to decide what we're going to do about Windows though. maybe it's an easy fix, not sure

nanaian:
If we can't easily fix this for Windows, perhaps we could just ignore the timeout values on Windows (i.e. default to 0s = no timeout)? If they are set, we can emit a warning that timeouts aren't supported on Windows.

ConorBobbleHat:
alright!
Despite my mentioning [wrapt_timeout_decorator](https://github.com/bitranox/wrapt_timeout_decorator) in the Discord, it turned out it was loosely typed enough to cause all sorts of headaches when trying to integrate it. Borrowing its secret sauce [dill](https://github.com/uqfoundation/dill) did allow for relatively painless Windows support, though.

PR #638, opened 2022-12-25
Add preliminary compiler options for Luigi's Mansion

Sage-of-Mirrors:
This PR adds a preset for Luigi's Mansion. However, I am babby, so the compiler options included are the bare minimum used while matching isolated functions. They will probably need to be updated once the full pool of options is found.


nanaian (approved):

ethteck:
We r babby 2

PR #651, opened 2023-01-07
Add PSY-Q 3.5 and PSY-Q 3.6 compilers for PlayStation 1

Xeeynamo:
# Overview

All the games released before 1997 for PlayStation 1 are most likely to use one of those two compilers. The PR adds support for them.

## Details

The PlayStation 1 compilers before PSY-Q 4.x are considerably different. They use MS-DOS/Win16 as development environment to run. Due to this limitation, the code I am proposing is going to install DOSEMU, a lightweight MS-DOS emulator that is easy to configure and to run command line programs to. I do not think the other alternative, DOSBox, is a good alternative as it struggles with CLI-only programs.

A limitation of DOSEMU is that paths needs to be in the 8.3 format (8 characters for the file name, maximum 3 characters for the extension). To not deal with MS-DOS naming mangling, I am copying the compilers in the same directory where the code to compile is (thanks @ethteck for the suggestion).

The PR also removes the old unused custom GCC 2.6.3 compiler as it was a work-around before the discover of DOSEMU.

## Technical limitations

DOSEMU needs `/proc/cpuinfo` to be available. Therefore I had to mount `/proc` read-only in the sandbox. Also DOSEMU wants the paths to be in the format `G:\object.oo`, therefore I could not use the variables `INPUT` and `OUTPUT` as they contains the `/tmp/` at their beginning. I had to burn in the code the name `object` unfortunately.

I was initially planning to add PSY-Q 3.3, but I found that the ASPSX.EXE of that SDK is generating an exception in DOSEMU. It is still downloading it, but for now I just excluded from the list of PlayStation 1 compilers in the wait for a patch.

## Extra

I am allowing edits by maintainers, so be free to push adjustments before reaching a merge-able state.

I noticed I accidentally submitted some formatted lines. Let me know if they are undesired and I will push a change to revert them.


ethteck (changes requested):
Looks good! I think we should remove the gcc string from compilers.json for the newly removed compiler. We'll also need a db migration to migrate the compiler id for all scratches that used to use the gcc263_mipsel compiler to the new psyq33 compiler

Xeeynamo:
I added the migration script 0030. Plus I marked the PR as ready to be reviewed!

backend/Dockerfile
```diff
@@ -61,16 +61,21 @@ ARG ENABLE_NDS_SUPPORT
...
     dpkg --add-architecture i386 && apt-get update && \
     apt-get install -y -o APT::Immediate-Configure=false \
-        wine; \
+    wine; \
     fi
 
 # ps1 specifics
 RUN if [ "${ENABLE_PS1_SUPPORT}" = "YES" ] ; then \
     apt-get install -y dos2unix; \
+    # all the following is required to run the MS-DOS toolchain
```
mkst, line 74: can this be `&&`'d together to ensure only 1 layer is created rather than 4? something like this (untested):

```
RUN if [ "${ENABLE_PS1_SUPPORT}" = "YES" ] ; then \
    apt-get install -y software-properties-common \
    && add-apt-repository -y ppa:dosemu2/ppa \
    && apt-get update \
    && apt-get install -y dos2unix dosemu2 \
fi
```

backend/coreapp/migrations/0030_replace_ps1_gcc263_with_psyq36.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    Migrate the compiler ID 'gcc-263-mipsel' into 'psyq3.6'
+    """
+
+    compiler_map = {
+        "gcc2.6.3-mipsel": "psyq3.6",
+    }
+
+    Scratch = apps.get_model("coreapp", "Scratch")
+    for row in Scratch.objects.only("compiler").filter(compiler__endswith="-psyq"):
```
mkst, line 17: is this the right filter clause?

backend/coreapp/compilers.py
```diff
@@ -263,13 +263,25 @@ def preset_from_name(name: str) -> Optional[Preset]:
 )
 
 # PS1
-GCC263_MIPSEL = GCCPS1Compiler(
-    id="gcc2.6.3-mipsel",
+PSYQ_MSDOS_CC = (
+    'cpp -P "$INPUT" | unix2dos > ${OUTPUT}c && cp ${COMPILER_DIR}/* . && '
```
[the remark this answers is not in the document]
Xeeynamo: Unfortunately I do not think I can. The parameter `-K .` mounts the current directory as `G:\` and there is no way for dosemu2 to have any visibility of what is outside that directory. That is also the reason why I hardcoded the name `object.o` rather than using `$OUTPUT`.

backend/coreapp/migrations/0030_replace_ps1_gcc263_with_psyq36.py
Xeeynamo: Thank you, I did not notice it.

backend/coreapp/sandbox.py
```diff
@@ -65,8 +65,10 @@ def sandbox_command(self, mounts: List[Path], env: Dict[str, str]) -> List[str]:
             "--bindmount_ro", "/lib",
             "--bindmount_ro", "/lib64",
             "--bindmount_ro", "/usr",
+            "--bindmount_ro", "/proc",
```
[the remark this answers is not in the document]
Xeeynamo: It does not and it seems to be valid. dosbox2 does not run if it is not able to fetch `/proc/cpuinfo`

backend/Dockerfile
```diff
@@ -70,7 +70,12 @@ RUN if [ "${ENABLE_NDS_SUPPORT}" = "YES" ] || \
 
 # ps1 specifics
 RUN if [ "${ENABLE_PS1_SUPPORT}" = "YES" ] ; then \
-    apt-get install -y dos2unix; \
+    apt-get install -y dos2unix && \
+    # all the following is required to run the MS-DOS toolchain
```
mkst, line 74: This comment will cause a new layer i think? I'd say just remove it

backend/coreapp/compilers.py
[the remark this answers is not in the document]
mkst: Ah yes, it's the path to the compiler rather than the input file that is the problem.

backend/coreapp/sandbox.py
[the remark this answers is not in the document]
mkst: I'm suggesting that the --disable-proc argument to the sandbox means that the bindmount of /proc doesn't happen

ethteck:
@Xeeynamo do you mind merging main into your branch? it has a conflict. also is this still in progress, or is it good to go after that?

mkst:
I'm not convinced this will work with docker+sandbox

Xeeynamo:
> @Xeeynamo do you mind merging main into your branch? it has a conflict. also is this still in progress, or is it good to go after that?

Done. I have been using this fork locally for a while and it works fine with `docker compose up` (cc @mkst). What stops me to get it merged are the failing tests, which I fail to understand how to fix. I would still like to see this merged down the line.

backend/Dockerfile
```diff
@@ -70,7 +70,11 @@ RUN if [ "${ENABLE_NDS_SUPPORT}" = "YES" ] || \
 
 # ps1 specifics
 RUN if [ "${ENABLE_PS1_SUPPORT}" = "YES" ] ; then \
-    apt-get install -y dos2unix; \
+    apt-get install -y dos2unix && \
+    apt-get install -y software-properties-common && \
+    add-apt-repository -y ppa:dosemu2/ppa && \
+    apt-get update && \
+    apt-get install -y dosemu2 \
```
mkst, line 77: needs to have a trailing `;` otherwise it's a syntax error.
```
    apt-get install -y dosemu2; \
```

mkst:
this syntax means we don't need to copy the compilers in:
```py
PSYQ_MSDOS_CC = (
    'cpp -P "$INPUT" | unix2dos > ${OUTPUT}c && ' +
    'TERM=xterm HOME=$(pwd) dosemu -f ${COMPILER_DIR}/.dosemurc -d ${COMPILER_DIR} -d $(pwd) -E "G:\\CC1PSX.EXE -quiet ${COMPILER_FLAGS} -o H:\\object.os H:\\object.oc" &&' +
    'TERM=xterm HOME=$(pwd) dosemu -f ${COMPILER_DIR}/.dosemurc -d ${COMPILER_DIR} -d $(pwd) -E "G:\\ASPSX.EXE  -quiet -o H:\\object.oo H:\\object.os" &&' +
    '${COMPILER_DIR}/psyq-obj-parser "${OUTPUT}"o -o "${OUTPUT}"'
)
```

the `-f .dosemurc` bit is optional, am struggling to make any meaningful configuration... the following doesnt seem to fix (m)any errors or warnings, nor speed things up :(
```conf
$_cpuspeed = (3000)

$_cpu_vm = "emulated"
$_cpu_vm_dpmi = "emulated"

$_hogthreshold = (0)

$_floppy_a = ""

$_lpt1 = ""
$_lpt2 = ""

$_speaker = ""
$_sound = (off)

$_pktdriver = (off)
$_ne2k = (off)

$_video = "none"

$_layout = "us"

$_hdimage = "+0 +1"

```

.github/workflows/ci.yml
```diff
@@ -54,8 +54,13 @@ jobs:
       - name: Install apt dependencies (cached)
         uses: awalsh128/cache-apt-pkgs-action@v1
         with:
-          packages: docker-ce docker-ce-cli containerd.io docker-compose-plugin binutils-aarch64-linux-gnu binutils-mips-linux-gnu binutils-powerpc-linux-gnu dos2unix libprotobuf-dev libnl-route-3-dev libncurses5 protobuf-compiler wine
+          packages: docker-ce docker-ce-cli containerd.io docker-compose-plugin binutils-aarch64-linux-gnu binutils-mips-linux-gnu binutils-powerpc-linux-gnu dos2unix libprotobuf-dev libnl-route-3-dev libncurses5 protobuf-compiler wine software-properties-common
           version: 1.0
+      - name: Install DOSEMU
+        run: |-
+          add-apt-repository -y ppa:dosemu2/ppa
```
ethteck, line 61: you need sudo in front of these commands, as github actions is complaining about

Xeeynamo:
@mkst I came up with the conclusion this is so far the best dosemu config, so far, in terms of performance:

```
$_cpu_vm = "emulated"
$_cpu_vm_dpmi = "emulated"
$_sound = (off)
$_hogthreshold = (0)

$_cpuspeed = (333.333)
$_pci = (off)
$_rdtsc = (off)
$_term_color = (off)
$_ipxsupport = (off)
$_pktdriver = (off)

```

The first four slashed the time execution by 4 (!!!) times and they are ordered in terms of performance impact. The rest have a negligible impact.

How did you found out about `_cpu_vm`? I do not see it documented anywhere!

ethteck:
everything here looks good to me, but docker tests are still failing for the new compilers - any idea why?

PR #656, opened 2023-01-10
A few things

ethteck:
Make logs more brief by hiding the PATH from compile commands 
Hopefully prevent the frontend server from making tons of anonymous profiles 
Add a scroll bar to the problems window
Restore the old look of ScoreBadge


nanaian (commented):
Merged already but LGTM :)

PR #659, opened 2023-01-19
Switch timeout mechanism to `subprocess.run`

ConorBobbleHat:
Okay: hopefully the last timeout-related PR from me for a while :)

This PR:
- Removes exception_on_timeout & its dependencies (i.e. dill) completely
- Adds a `timeout` argument to run_subprocess that passes through to `subprocess.run`
- Remove the timeout guard on m2c calls (as it's currently called as a module, not a subprocess)
- Adds a new compiler, `DummyLongRunningCompiler` that does nothing but wait for an hour if invoked
- Adds a new test, `test_compiler_timeout`, that sets a compilation timeout of three seconds, invokes  `DummyLongRunningCompiler`, and verifies a timeout error was generated
- Removes the time limits passed to nsjail (to prevent race conditions from having us have to handle two different variants of timeout errors)

Tested on both django's `runserver` and gunicorn, and verified that that a) both work happily with this timeout mechanism, and b) the impact on latency of compile calls is negligible (within the range of network jitter).

(with many thanks to @simonlindholm for the methodology)


nanaian (commented):
These changes all look good! Glad to simplify, and I like the new test. Thanks ^-^

Would you be able to add a test or otherwise verify that COMPILATION_TIMEOUT_SECONDS=0 results in no timeout being applied (i.e. equivalent to passing `timeout=None` to `subprocess.run`)?

ethteck (approved):

ConorBobbleHat:
A timeout of zero should now disable timeouts entirely

PR #661, opened 2023-01-22
Add PS2 EE GCC 2.9 and 3.2 compilers

entriphy:
This PR adds two compilers for PS2 (binaries are not included in this PR):
* `ee-gcc2.9-990721` (required to make this decomp match: https://decomp.me/scratch/z3eYN)
* `ee-gcc3.2-040921`

I added the build number/date to the ID since a slightly newer 2.9 compiler exists (`2.9-991111`), although that compiler seems to be broken, so it's moreso in the event that another EE compiler is added in the future with the same version but different build number (I didn't change the ID for `ee-gcc2.96` since I wasn't sure if that would break stuff on prod).


ethteck:
Thanks for this! As part of adding new compilers, we typically request people modify download.py such that it downloads them.  If this isn't feasible, feel free to contact me and we can find out how to get them on the live site. Are you in our discord?

PR #662, opened 2023-01-23
Change egcs defaults and add two presets

AngheloAlf:
- We figured out a few better presets for egcs which we believe will be enabled for almost every project that uses this specific compiler
- This also adds a preset for the iQue version of Dr. Mario 64 and libultra


nanaian:
Applying default flags might retroactively change scratches- unsure if its a safe change

ethteck:
I think the flags changed are not going to adversely affect anything, and it's also a relatively unused compiler, so we should be okay

PR #668, opened 2023-02-01
Update LoK: SR compiler version and flags

Gh0stBlade:
The correct compiler version for KAIN2.EXE is "mostly" PsyQ 4.3.

The AADLIB is what uses 4.5 I think so it's probably best to use 4.3 since 80% of the code is using that.

### DIFF
--- a/backend/coreapp/compilers.py
+++ b/backend/coreapp/compilers.py
@@ -898,8 +898,8 @@ _all_presets = [
     ),
     Preset(
         "Legacy of Kain: Soul Reaver",
-        PSYQ45,
-        "-g -Wall -O2 -G256",
+        PSYQ43,
+        "-O2 -G65536",
     ),
     Preset(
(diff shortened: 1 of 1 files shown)


nanaian (approved):
can ignore failing tests- changes look good.
(not sure what's going on with the tests. vercel failed because of the live API being faulty somehow, black seems to have failed due to a versioning problem, and backend tests are a compiler download issue)

ethteck:
> the live API being faulty somehow 
that's concerning - what do you mean?

I fixed the black thing, and the docker test is a known issue and has to do with permissions of one of the compilers. I'm re-running the failed tests just to make sure black passes.

nanaian:
[looks like the API responded with html](https://vercel.com/decompme/frontend/9DvRPt47xP96jpUJjBUB1myNPryi#L55)

PR #674, opened 2023-02-07
Adding IDO 6.0

LLONSIT:


nanaian:
black is failing

LLONSIT:
Qemu-irix with binfmt seems to be working

ethteck:
To echo what I said in the discord - black just wanted you to add a space after the # in your comment, so no need to delete the line entirely.

Additionally, we can't accept this PR as-is because the setting up of the compiler needs to happen in download.py rather than in the ci build script. You'll need to migrate the setup steps so any consumer of the repo can properly run the compiler.

LLONSIT:
Just need the qemu-irix command for the docker

.github/workflows/ci.yml
```diff
@@ -162,9 +175,11 @@ jobs:
         run: |-
           mkdir -p sandbox && chmod 777 sandbox
           mkdir -p local_files && chmod 777 local_files
+          #ls /ido_root
```
ethteck, line 178: could you remove this debugging line?

.github/workflows/ci.yml
```diff
@@ -174,7 +189,9 @@ jobs:
             -e USE_SANDBOX_JAIL=on \
             -e SANDBOX_DISABLE_PROC=true \
             -e TIMEOUT_SCALE_FACTOR=10 \
-            decompme_backend -c 'cd /decomp.me/backend && \
+            -e QEMU_LD_PREFIX=/ido_root \
+            -e RUNNING_IN_DOCKER=1\
```
ethteck, line 193: hard-coding this to 1 here seems wrong

.github/workflows/ci.yml
```diff
@@ -174,7 +189,9 @@ jobs:
...
+            decompme_backend -c 'echo $QEMU_LD_PREFIX && echo $RUNNING_IN_DOCKER && cd /decomp.me/backend && \
```
ethteck, line 194: I think this is a debugging line and can be removed

ethteck (changes requested):
Some more thing I found. @mkst @EllipticEllipsis could you two also review the most recent state of this, since I think you both have useful input probably (re: docker, re: changing the default flag for IDO 5.3 Pascal, respectively, along with anything else you see)

LLONSIT (commented):
Ok

EllipticEllipsis:
Given that most of 5.3 seems to use -mips1, it's probably sensible to change it (although I wish it had been mentioned in the PR description).

Could you try and get IDO 6.0 working with a fork of recomp? It shouldn't be too difficult if it's O32 and you don't need many manual hacks (most of them are for `copt`, which hopefully you don't need for this!). And it'd be a lot more convenient than qemu if it does work.

ethteck:
As mentioned in Discord, I think the way to go forward here is what EllipticEllipsis said above. I don't think adding all these dependencies and new requirements to CI / docker / any new host is a good idea for these compilers which will realistically only ever be used on the site by one or two people. If you're willing to put some effort into debugging the issues with ido recomp, I'm sure we can figure out a way to get these binaries working under that process. Until then, I'm going to close this, since I don't think it's appropriate for the site in its current form.

Please do feel free to raise another PR with your new AeroGauge JP Kiosk Demo preset and change to IDO 5.3 Pascal compiler flags if you'd like in the meantime.

PR #690, opened 2023-02-23
Add django-session-timeout

### LINKED ISSUE #689: The site seems to log people out
why? let the mystery...begin!

ConorBobbleHat:
Fixes #689.

As per https://discord.com/channels/897066363951128586/897075447534354462/1078353419032592524 : add `django-session-timeout` (https://github.com/labd/django-session-timeout/) and, configure it to only revoke sessions if they've been inactive for ninety days.


backend/decompme/settings.py
```diff
@@ -221,6 +225,10 @@
 SENTRY_DSN = env("SENTRY_DSN", str)
 SENTRY_SAMPLE_RATE = env("SENTRY_SAMPLE_RATE", float)
 
+SESSION_COOKIE_AGE = env("SESSION_COOKIE_AGE", int)
```
nanaian, line 228: Can you add a large default for this in .env.example? It looks like the default of the library will be only 3600s

ConorBobbleHat: 👍; mirrored the default into .env

PR #702, opened 2023-03-12
IDO5.3 C++

LLONSIT-glitch:
(Reopened again)

Introduce a "C++ translator", which is made specifically for IDO5.3 (date 11/94 same as ido5.3 pascal)


LLONSIT (approved):
Ap

PR #706, opened 2023-03-20
5.3 C++

LLONSIT:
Implement the ido5.3 c++ translator


mkst:
Can we set some env var or flag so we don't have to mount a writeable directory from the host?

simonlindholm:
or binding it to the same thing as `/tmp`?

simonlindholm:
No, don't bind /tmp the same broken way. /tmp is already bound! Just do the same thing for /var/tmp as is already done for it.

LLONSIT:
> No, don't bind /tmp the same broken way. /tmp is already bound! Just do the same thing for /var/tmp as is already done for it.

"--bindmount", f"{self.path}:/var/tmp" ?

simonlindholm:
yeah

PR #707, opened 2023-03-22
Make families larger

ethteck:
allow families to include scratches with differing compilers


nanaian:
Worth bearing in mind that this will make the 'someone else matched this function!' banner appear when the "match" was actually on a different compiler - unsure if we want that

PR #728, opened 2023-04-07
Initial Saturn support

sozud:
This gets some basic Saturn support working. I needed to update to Ubuntu 22.04 to get binutils-sh-elf, which doesn't seem to be in 20.04. An example to test with is:

`Compile with -O0`

```
mov.l   r14,@-r15
sts.l   pr,@-r15
add     #-4,r15
mov     r15,r14
mov.l   r4,@r14
mov.l   @r14,r0
cmp/eq  #1,r0
bf      16
mov     #0,r0
bra     38
nop
mov.l   @r14,r7
mov     r7,r4
mov     #10,r5
mov.l   44,r6
jsr     @r6
nop
mov     r0,r2
mov     r2,r1
shll2   r1
add     r0,r1
mov     r1,r2
add     r2,r2
mov     r7,r1
sub     r2,r1
mov     r1,r0
bra     38
nop
add     #4,r14
mov     r14,r15
lds.l   @r15+,pr
mov.l   @r15+,r14
rts
nop
.word 0x0000
```

```
int test(int arg0)
{
    if (arg0 == 1)
    {
        return 0;
    }

    return arg0 % 10;
}
```

This example doesn't totally match since I think I had the compiler settings different for the generated assembly. We don't have splitting tools for Saturn yet so this work is still somewhat speculative. I suspect this is the compiler used for Saturn SOTN but I'm not sure.

I needed to wrap the executables in batch files because they expect a particular PATH environment to work.


ethteck:
Exciting! Sucks we need dosbox though :/ and wondering why you bumped python to 3.10 - is that just the default for the newer ubuntu?

sozud:
As far as I can tell there is no Python3.9 package in 22.04. https://packages.ubuntu.com/search?suite=jammy&section=all&arch=any&keywords=python3.9&searchon=names

I just tried running the compiler in Wine and get this:
```
winevdm: Z:\GCCSH\BIN\CC1.EXE is a DOS application, you need to install DOSBox.
```
I went with dosemu since it's the solution that that PS1 DOS compilers are using.

ethteck:
Makes sense. Could you fix the black errors / non-docker build as well (.github/workfflows/ci.yml)? And it looks like the docker test is failing too...something with the sandbox. have you run it locally with the sandbox?

sozud:
So running the full CI requires your approval? I haven't tried running tests locally, I haven't been able to run this outside docker due to issues with poetry being unable to finish an install.

ethteck:
It requires my approval since you're a new contributor, yes. What's the issue with poetry?

ethteck:
the setup python thing failed trying "3.1"??? I guess just keep it at 3.9 for now

sozud:
On my Linux VM poetry gets stuck forever trying to install. I think it's this issue: https://github.com/python-poetry/poetry/issues/3352
I've tried using the old installer and removing the cache but it still gets stuck.

On my Mac I run into this
```
  Hash for rapidfuzz (2.13.7) from archive rapidfuzz-2.13.7-cp310-cp310-macosx_10_9_x86_64.whl not found in known hashes (was: sha256:3d3843b66a57d566b2a4c0ed85c4d87b6d46aef8e5116fb08a89f0c6fb737ab5)
```

I think the issue with the docker tests is an invalid argument: `'sh-elf-objdump --disassemble --disassemble-zeroes --line-numbers --reloc --start-address=0 -m mips:4300 /tmp/out.s'`. The mips part shouldn't be there. I'll see if I can figure out a workflow to debug this.

Ced2911:
could it be because of this ? https://github.com/sozud/decomp.me/blob/saturn/backend/coreapp/diff_wrapper.py#L231

.github/workflows/ci.yml
```diff
@@ -13,10 +13,10 @@ jobs:
 
       - name: Install poetry
         run: pipx install poetry
-      - name: Setup Python 3.10
+      - name: Setup Python 3.10.0
         uses: actions/setup-python@v4
         with:
-          python-version: 3.10
```
nanaian: lol yaml, probably better to solve this with "3.10" as an explicit string

sozud:
I think I figured out the issue, somehow I lost the change with the asm-differ hash update. Could someone approve the workflow?

sozud:
It looks like everything is passing now, are any other changes needed?

PR #730, opened 2023-04-08
Hide projects feature

ethteck:
since it's bugged for now and will undergo a revamp soon :tm:


nanaian (approved):
see also my remove-projects branch

PR #732, opened 2023-04-09
Fix saturn path issue and use objcopy to convert to elf

sozud:
This changes the directory layout of the compiler so that the batch files and absolute paths aren't needed. This also includes the change by @Ced2911 to use objcopy to convert the coff to elf. I don't understand why this is necessary for some inputs on certain systems but it seems to work fine locally.

### DIFF
--- a/backend/compilers/download.py
+++ b/backend/compilers/download.py
@@ -613,7 +613,10 @@ def download_saturn():
         url="https://github.com/sozud/saturn-compilers/archive/refs/heads/main.zip",
     )
 
-    shutil.move(f"{COMPILERS_DIR}/saturn-compilers-main/cygnus-2.7-96Q3", COMPILERS_DIR)
+    shutil.move(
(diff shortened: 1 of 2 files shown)


ethteck:
What about the scripts that are used to invoke the compiler? Those still have hard-coded drive letters, right?

sozud:
This now uses this folder https://github.com/sozud/saturn-compilers/tree/main/cygnus-2.7-96Q3-bin
We just copy that over
```
    shutil.move(
        f"{COMPILERS_DIR}/saturn-compilers-main/cygnus-2.7-96Q3-bin",
        f"{COMPILERS_DIR}/cygnus-2.7-96Q3",
    )
```
Since everything is in the same directory it's basically like it's all on the PATH. I thought the headers and crt0 etc. were required but it doesn't seem like it for the way decomp.me works.

Issue #639, opened 2022-12-29
Language server


nanaian:
- https://www.npmjs.com/package/codemirror-languageserver
- clangd as a service

PR #739, opened 2023-05-02
Add clangd language server support

ConorBobbleHat:
(woo!)
(#639)

This draft contains a very bare-bones first pass at integrating support of clangd compiled to wasm.

I've taken a page out of [ffmpeg-wasm](https://github.com/ffmpegwasm/ffmpeg.wasm)'s book, and I've created a package on npm ([@clangd-wasm/clangd-wasm](https://www.npmjs.com/package/@clangd-wasm/clangd-wasm)) that exports a transport compatible with `codemirror-languageserver`. At runtime, the client then pulls in the compiled wasm binary from unpkg. Fingers crossed, this makes life easier then manually shifting things around into the `public` folder.

Right now, this enables clangd diagnostics and hover tooltips in both the scratch and context editor.

### Language server integration
- [x] ~For some reason, `codemirror-languageserver` won't load when imported as an external dependency. I've gotten it to work successfully in a fresh webpack app, but haven't been able to get it to work in next (i.e. decomp.me). For the sake of getting something to work, this PR currently houses an in-tree copy of its sources.~ Fixed, including sources in tree as fork.
- [x] ~Although the source and context editors both receive completion separately, clangd doesn't recognize them as part of the same project. As such, the source editor can't see any definitions from the scratch.~ Fixed by implementing save notifications, and supplying a compile_commands that has the source reference the context.
- [x] ~Right now, we presume everything is C++. Predictably, this causes issues.~ Discrimination between C / C++ implemented.
- [x] ~`codemirror-languageserver` pulls in a dependency of some kind that causes webpack to throw a funky warning about not being able to resolve modules. It doesn't seem to affect the success of the build, but is probably worth further investigation.~ Fixed by telling webpack everything's going to be okay.
- [x] ~`codemirror-languageserver` currently doesn't support requesting document formatting, even though clangd supports it through `clang-format` and `clang-tidy`~ Implemented on Alt-Shift-F. TODO: add button to toolbar? Expose clang-tidy options on a preset / user basis?
- [x] ~clangd emits diagnostics with newlines that codemirror collapses to a single line, which can be confusing.~ Implemented.

### WebAssemblyification
- [x] ~I haven't done much work to minimize the size of the wasm binary at all - it's currently 50MB uncompressed, and about 15MB compressed. That being said, that means there's still probably some low hanging fruit: using `-Os` instead of `-O3`, using [asyncify's optimization flags](https://emscripten.org/docs/porting/asyncify.html#optimizing), etc.~
- [x] ~a 50MB executable that needs to be duplicated across worker threads probably means that the language server has a non-trivial memory footprint. Trying to get an estimate of how much proved difficult - I was getting measurements all over the place. If anyone has expertise in this area, I'd really appreciate some help quantifying this.~
- [x] ~clangd seems to support command line options to limit the number of threads used - worth investigating?~
All duly investigated; small version added.

### UX
- [x] ~Right now, the language server is downloaded automatically in the background. This should absolutely be gated behind an option of some sort. [The discussion on discord](https://discord.com/channels/897066363951128586/897075447534354462/1087878053314703402) gravitated towards a more generic download manager system with an eye for the future.~ Implemented as a checkbox under editor options.

#### Links
- core module: https://github.com/ConorBobbleHat/clangd-wasm-core (build scripts to produce the wasm binary)
- wrapper module: https://github.com/ConorBobbleHat/clangd-wasm


ConorBobbleHat:
forgot to add in the above:
- [ ] cross origin isolation breaks github login popup - switching to redirects will fix this

nanaian:
> Right now, the language server is downloaded automatically in the background. This should absolutely be gated behind an option of some sort. [The discussion on discord](https://discord.com/channels/897066363951128586/897075447534354462/1087878053314703402) gravitated towards a more generic download manager system with an eye for the future.

Some kind of UI wrapper that basically just does a bunch of dynamic imports would be good I think. We can have the modules export specific funcs depending on what the module wants to do (e.g. clangd lsp needs to hook into codemirror)

nanaian:
>  Right now, we presume everything is C++. Predictably, this causes issues.

We can ask the compiler of the scratch what language it is. See `useCompilers`

nanaian:
> I haven't done much work to minimize the size of the wasm binary at all - it's currently 50MB uncompressed, and about 15MB compressed. That being said, that means there's still probably some low hanging fruit: using -Os instead of -O3, using [asyncify's optimization flags](https://emscripten.org/docs/porting/asyncify.html#optimizing), etc.

Can also use wasm-opt and strip as second passes

nanaian:
Based on all the issues with codemirror-languageserver I think'll we'll want to vendor our own fork of it anyway. Can contribute things upstream but they might not accept them

frontend/src/components/Scratch/Scratch.tsx
```diff
@@ -146,6 +148,44 @@ export default function Scratch({
     const sourceCompareExtension = useCompareExtension(sourceEditor, shouldCompare ? parentScratch?.source_code : undefined)
     const contextCompareExtension = useCompareExtension(contextEditor, shouldCompare ? parentScratch?.context : undefined)
 
+    useEffect(() => {
+        const initLanguageServer = async () => {
+            // TODO: make loading this conditional on user opt-in
+            const { ClangdStdioTransport } = await import ("@clangd-wasm/clangd-wasm")
```
nanaian, line 154:
```suggestion
            const { ClangdStdioTransport } = await import("@clangd-wasm/clangd-wasm")
```

nanaian (changes requested):

nanaian:
> > Right now, the language server is downloaded automatically in the background. This should absolutely be gated behind an option of some sort. [The discussion on discord](https://discord.com/channels/897066363951128586/897075447534354462/1087878053314703402) gravitated towards a more generic download manager system with an eye for the future.
> 
> Some kind of UI wrapper that basically just does a bunch of dynamic imports would be good I think. We can have the modules export specific funcs depending on what the module wants to do (e.g. clangd lsp needs to hook into codemirror)

We're gonna need this for #740 anyway, so worth implementing both at once. Maybe can do a separate PR for that just has a dummy option for language server

ConorBobbleHat:
> > Some kind of UI wrapper that basically just does a bunch of dynamic imports would be good I think. We can have the modules export specific funcs depending on what the module wants to do (e.g. clangd lsp needs to hook into codemirror)
> 
> We're gonna need this for #740 anyway, so worth implementing both at once. Maybe can do a separate PR for that just has a dummy option for language server

I like the approach! I'm unclear as to how exactly that'd be architected, though. Would it be specifically a UI module system, with a fixed interface modules need to adhere to in order to override a handful of specific behaviours, or a more general plugin system to allow near-arbitrary parts of the site and codebase to be toggled on and off? (or something else entirely again?)

ConorBobbleHat:
> > 
> 
> Can also use wasm-opt and strip as second passes. And brotli-compress?

I'll need to double check - I think emscripten integrates (some but maybe not all of?) the same binaryen optimization passes as wasm-opt and exposes them as the standard set of optimization flags. Stripping and brotli-compression are definitely good shouts, though.

ConorBobbleHat:
> Based on all the issues with codemirror-languageserver I think'll we'll want to vendor our own fork of it anyway. Can contribute things upstream but they might not accept them

That's fair - it's looking like the language server integration might get more than a little bespoke.


> I think'll we'll want to vendor our own fork of it anyway

When you say fork - are you thinking a github style, out of tree fork? If so, what's the best way of pulling that off?

ConorBobbleHat:
> > 
> 
> We can ask the compiler of the scratch what language it is. See `useCompilers`

I may well be looking in the wrong place - I can't see an obvious way to pull a scratch's language out based on just the data from `useCompilers`? Is there a standard flag or set of flags that can be parsed out?

nanaian:
My bad, looks like https://decomp.me/api/compilers doesn't surface `language` - the backend has the data we just need to write it out

ethteck:
Language and compiler isn't a 1:1 mapping. You can use a Lang flag on many compilers to change the input language

nanaian:
> Language and compiler isn't a 1:1 mapping. You can use a Lang flag on many compilers to change the input language

mb, do we expose language on /api/scratch?

nanaian:
> When you say fork - are you thinking a github style, out of tree fork? If so, what's the best way of pulling that off?

Nah, let's keep it where you have it for now. Perhaps later we can make this repo into a [monorepo](https://yarnpkg.com/features/workspaces) and publish it on npm

ConorBobbleHat:
sorry - my bad; not intentional. You got it.

frontend/src/lib/codemirror/languageServer.ts
```diff
@@ -411,6 +419,18 @@ class LanguageServerPlugin implements PluginValue {
         }
     }
 
+    async requestFormat() {
+        return await this.client.textDocumentFormatting({
+            textDocument: {
+                uri: this.documentUri,
+            },
+            options: {
```
nanaian, line 427: I presume clangd uses clang-format - do we know how to provide full clang-format options? Do we need to create a `.clang-format` vfs file or something?

ConorBobbleHat: Yup; that's exactly it. I've added support to clangd-wasm to create `compile-commands.json` somewhat elegantly; it probably makes sense to treat `.clang-format` the same way.

to pick your brain: I'm going back on forth on the best way to expose the format settings. A "decomp.me profile" that's baked into the site? Preset-specific settings? Just exposing the whole thing to end users? (get an LLM to guess based on context and the weather? :))

nanaian: I think exposing the whole thing to end users is fine, many projects (presets) will have an existing clang-format config that they would do well to just copy paste in. We can probably format config this for a later enhancement and just use a sensible default for now (make an issue?) unless you want to tackle it

nanaian: I think a ClangFormat model that just holds the yaml as a string would be fine, and we can make CompilerConfig (#614) optionally reference one. If there is no ClangFormat referenced we can use a decomp.me default or a user setting default

ConorBobbleHat: As a methodology; that seems pretty sound!
There's enough left on the todo list that unless someone wants to vouch for this as critical, I'll probably move it onto the future brush-up pile. I'll wire up a default `.clang-format` for this PR, which should tee things up nicely for future work.

nanaian: pmret/papermario has a .clang-format that we like that you could use

nanaian: oh nice, you already did lol

frontend/next.config.js
```diff
@@ -111,6 +111,13 @@ let app = withPlausibleProxy({
             },
         })
 
+        // This lets us import yaml files, and have their contents
+        // be baked into the output bundle as a static string
+        config.module.rules.push({
+            "test": /\.yaml/,
+            "type": "asset/source",
+        })
```
nanaian, line 119: I feel like it might be better to `fetch` the file as text and use `import.meta.url`

nanaian: I initially expected the `yaml` import to parse the data like importing `json` does

nanaian:
> Right now, the language server is downloaded automatically in the background. This should absolutely be gated behind an option of some sort. [The discussion on discord](https://discord.com/channels/897066363951128586/897075447534354462/1087878053314703402) gravitated towards a more generic download manager system with an eye for the future.

I think a settings checkbox with a note that says 'warning this is a bigly download' will be fine, actually. We can chuck a loading spinner next to it while the download goes on. I don't think we really need much new UI for this

frontend/src/components/Scratch/hooks/useLanguageServer.ts
```diff
@@ -83,11 +97,30 @@ export default function useLanguageServer(scratch: api.Scratch, sourceEditor: Mu
...
+            if (saveSource)
+                await saveSource(scratch.source_code)
+        })()
+    }
+
+    const saveContextRet = () => {
+        (async () => {
+            if (saveContext)
+                await saveContext(scratch.context)
+        })()
```
nanaian, line 122: This will have the same behaviour as just calling `saveContext(scratch.context)` and discarding the Promise, no?

ConorBobbleHat: Oops; you're right. I think that's the result of me refactoring out some more complicated behavior, and then forgetting to unwrap that.

frontend/next.config.js
ConorBobbleHat: I do like the slickness of having it be transparently included in the bundle, but I can see how importing structured data and getting a string back is maybe not the best way for that to behave. Will change.

ConorBobbleHat:
> > 
> 
> I think a settings page checkbox with a note that says 'warning this is a bigly download' will be fine, actually. We can chuck a loading spinner next to it while the download goes on. I don't think we really need much new UI for this

No objections here - UI work is * not * my forte :)

frontend/next.config.js
ConorBobbleHat: As it turns out, doing `new URL("./default-clang-format.yaml", import.meta.url).toString()` evaluates to the file contents anyway? Slightly confused, but it does keep things simple.

ConorBobbleHat: ... and that's stopped working.
Switching to fetch.

frontend/src/app/(navfooter)/settings/editor/EditorSettings.tsx
```diff
@@ -32,5 +51,15 @@ export default function EditorSettings() {
                 </div>
             </Checkbox>
         </Section>
+        <Section title="Language server">
+            <Checkbox
+                checked={languageServerEnabled}
+                onChange={maybeDownloadLanguageServer}
+                label="Enable in-browser language server"
+                description="Enable editor features such as code completion, error checking, and formatting via clangd and web assembly magic. WARNING: enabling will prompt a one time ~15MB download, and bump up resource usage during editing.">
```
nanaian, line 59:
```suggestion
                description="Enable editor features such as code completion, error checking, and formatting via clangd and WebAssembly magic. WARNING: enabling will prompt a one time ~15MB download, and bump up resource usage during editing.">
```

frontend/src/app/(navfooter)/settings/editor/EditorSettings.tsx
```diff
@@ -32,5 +51,15 @@ export default function EditorSettings() {
                 </div>
             </Checkbox>
         </Section>
+        <Section title="Language server">
+            <Checkbox
+                checked={languageServerEnabled}
+                onChange={maybeDownloadLanguageServer}
+                label="Enable in-browser language server"
```
nanaian, line 58:
```suggestion
                label="Enable language server"
```

nanaian (changes requested):
LGTM otherwise. Make sure to update "~ 15 MB" to whatever the actual size of the download is after stripping/optimising/etc

ConorBobbleHat:
Copy changed; will do on the WASM file size stat.

ConorBobbleHat:
> Stripping and brotli-compression are definitely good shouts, though.

It got slightly lost in the cracks, but to make this explicit: it turns out gzip performs pretty much exactly as well on WebAssembly as brotli, and unpkg has it on for everything it serves by default (hence no explicit compression anywhere in the codebase)

nanaian (approved):

frontend/src/app/(navfooter)/settings/editor/EditorSettings.tsx
```diff
@@ -32,5 +54,27 @@ export default function EditorSettings() {
...
+                <Checkbox
+                    checked={preferSmallLanguageServer}
+                    onChange={state => {
+                        setPreferSmallLanguageServer(state)
+                        setLanguageServerSettingsMutated(true)
+                    }}
+                    label="Prefer small language server build"
+                    description="Use a smaller (11MB vs 13MB download) build of the language server, at the expense of some performance."/>
+
+                {downloadingLanguageServer && <LoadingSpinner width="24px" />}
```
nanaian, line 76: Can you write "Loading" or something next to this? e.g.
```
<div className="pa-4"><LoadingSpinner /> Downloading...</div>
```

frontend/src/app/(navfooter)/settings/editor/EditorSettings.tsx
```diff
@@ -32,5 +54,27 @@ export default function EditorSettings() {
                 </div>
             </Checkbox>
         </Section>
+        <Section title="Language server">
+            <Checkbox
+                checked={languageServerEnabled}
+                onChange={state => {
+                    setLanguageServerEnabled(state)
+                    setLanguageServerSettingsMutated(true)
```
nanaian, line 62: Perhaps this could be set with an effect?

nanaian (changes requested):
Couple comments.
Also I'm trying to test on https://frontend-ajagvia1b-decompme.vercel.app/ after enabling the setting but it seems to not be working? Are there keybindings I need to know or is it not working on the Vercel preview?

ConorBobbleHat:
> Also I'm trying to test on https://frontend-ajagvia1b-decompme.vercel.app/ after enabling the setting but it seems to not be working?

That'd be it trying to be smart and seeing which language to use, but the live backend not returning a language field.
For the sake of review, I'll hardcode things so it's always on in C mode.

nanaian:
> For the sake of review, I'll hardcode things so it's always on in C mode.

Ah its OK ill just run locally :+1:

nanaian:
Awesome, it works!

Can you make the font size of popups smaller? e.g. use the same font and font size as the editor
image

frontend/src/app/(navfooter)/settings/editor/EditorSettings.tsx
ConorBobbleHat: I think that could run into the same problems with running on mount, but I was able to use an effect to make things read less clunkily.

frontend/src/app/(navfooter)/settings/editor/EditorSettings.tsx
ConorBobbleHat: Done

frontend/src/app/(navfooter)/settings/editor/EditorSettings.module.scss
```diff
@@ -0,0 +1,5 @@
+.loading {
+    display: flex;
+    gap: 0.5em;
+    padding: 1em;
+}
```
nanaian, line 5: Probably easier to use Tailwind `flex gap-2 p-4`

ConorBobbleHat: fair 'nough

ConorBobbleHat:
image

?

ConorBobbleHat:
> Can you make the font size of popups smaller? e.g. use the same font and font size as the editor. And add a few px of padding

checked in.



> Sorry for all the requests 😅

np! that's what review's for :)

ethteck:
Since I ended up writing a lot on discord, I'll summarize my feedback here:

* I would prefer if we disable the tab aligning feature in the formatter options
* It's unclear how to "use" the language server. I had to read the code to figure out the shortcut for auto-formatting
  * It'd be nice to have a Format button next to Compile / Decompile if the option is checked
  * This further demonstrates the need for some shortcut editor / viewer (later work)
* I'm only getting any feedback from the language server in the form of warning squiggles on function args, at least for the test function I have locally. I tried introducing compiler errors and I don't get any feedback. I'm not sure why this is or how I'd go about debugging it..

frontend/src/components/Scratch/hooks/default-clang-format.yaml
```diff
@@ -0,0 +1,52 @@
...
+  PadOperators: true
+AlignConsecutiveDeclarations: # structs etc
+  Enabled: true
+  AcrossEmptyLines: false
+  AcrossComments: false
+AlignConsecutiveMacros:
+  Enabled: true
+  AcrossEmptyLines: false
+  AcrossComments: false
+AlignTrailingComments: true
```
ethteck, line 47: I think some / all of these are the ones I'd prefer to just remove (assuming the default is to not align)

nanaian: you want `AlignConsecutiveAssignments` off I think

nanaian: will need to carry these over to papermario too; discuss with clover

nanaian (approved):
ty both, apologies for requesting the small/large option and then backtracking!

PR #741, opened 2023-05-04
Tiebreak family sorting by score using Last Modified

EllipticEllipsis:
Also reverse sort of Last Modified to have most recent first, I think this makes more sense overall since people generally want the more recent ones, so they should go in the consistent screen position, at the top.


EllipticEllipsis:
Also since currently the family pane cannot scroll, if there's too many the most recently modified will simply be inaccessible with the current ordering.

frontend/src/components/Scratch/SortableFamilyList.tsx
```diff
@@ -37,6 +37,10 @@ function compareScratchScores(a: TerseScratch, b: TerseScratch) {
     const aScore = a.score < 0 ? Infinity : a.score
     const bScore = b.score < 0 ? Infinity : b.score
 
+    // Sort scratches with the same score with most recently updated first
+    if (aScore == bScore) {
```
ethteck, line 41: Can we not do `b.last_updated` - `a.last_updated` ?

nanaian (approved):
LGTM

PR #749, opened 2023-05-09
Migrate to metadata api from head.tsx / manually setting title

ethteck:
Can shomeone h3lp


marijnvdwerf:
- `PageTitle.tsx` can probably be removed now
- For the ' | decomp.me' part we can probably use a template: https://nextjs.org/docs/app/api-reference/functions/generate-metadata#template
- The old version of `PageTitle` set the title, description, `og:site_name`, `og:title` and `og:description` tags. The first of those are handled, but what about the `og:` tags? Are they generated with title/description, or are they not generated? Do we even need them?
- `darkreader-lock` is being declared in multiple places, we might not need to worry about this.

| | `<PageTitle>` | generateMetadata |
|--------|--------|--------|
| `title` | `title` | ✅ |
| `meta[name=description]` | `description` | ✅ |
| `meta[property="og:site_name"]` | decomp.me |   |
| `meta[property="og:title"]` | `title` |   |
| `meta[property="og:description"]` | `description` |   | 
| `meta[name=darkreader-lock]` | ✅ |   |

nanaian:
Ah, that looks fine then RE: darkreader-lock. For opengraph we should endeavour to output the correct tags so we get nice social embeds on Facebook, Discord, etc.

PR #751, opened 2023-05-10
Remove saturn directories when downloading

sozud:
I tested this by running the download.py script multiple times outside the container. I think this should fix situations where saturn-compilers-main is out of date, or cygnus-2.7-96Q3 already exists.


ethteck:
thank you!

PR #769, opened 2023-05-18
Fix switching diff algorithms

### LINKED ISSUE #768: Issues swapping between diff algorithms after changing it for the first time
After changing the diff algorithm, it starts to behave weirdly when trying to change it a 2nd time
This is on firefox 113.0.1
https://gyazo.com/66e5115dd98a76707c9d2feca68b1e2c

ConorBobbleHat:
Fixes #768.

Turns out both flags were being set at once, and given that Quantum Diffing (™️) still isn't stable, this PR fixes that.

### DIFF
--- a/frontend/src/components/compiler/CompilerOpts.tsx
+++ b/frontend/src/components/compiler/CompilerOpts.tsx
@@ -15,6 +15,7 @@ const NO_TRANSLATION = "NO_TRANSLATION"
 interface IOptsContext {
     checkFlag(flag: string): boolean
     setFlag(flag: string, value: boolean): void
+    setFlags(edits: {flag: string, value: boolean}[]): void
 }
 
 const OptsContext = createContext<IOptsContext>(undefined)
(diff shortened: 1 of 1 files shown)


ethteck (approved):

frontend/src/components/compiler/CompilerOpts.tsx
```diff
@@ -63,17 +64,19 @@ function FlagSet({ name, children, value }) {
...
         <Select
             onChange={event => {
-                for (const child of children) {
-                    setFlag(child.props.flag, false)
-                }
+                const trueFlag = (event.target as HTMLSelectElement).value
 
-                setFlag((event.target as HTMLSelectElement).value, true)
+                const edits = children.map(child => {
+                    return { flag: child.props.flag, value: child.props.flag == trueFlag }
```
nanaian, line 76: I'm hesistant here because `DiffFlagSet`'s `children` arg isn't typed, but I guess pretty much none of this file is typed so its fine?

nanaian (commented):
LGTM if we dont want to add typing

nanaian (approved):
ty!

ConorBobbleHat:
np!

Issue #757, opened 2023-05-16
Branch indicators lack contrast


marijnvdwerf:
In light mode, most branch indicators don't contrast enough with the background.

Also, the colours seem pretty basic (`#fof` and `#0ff`) and all over the place, it might be interesting to look into adopting some predefined palette. Some common datavis palettes can be found at https://plotly.com/python/discrete-color/#color-sequences-in-plotly-express.

Example:
<img width="291" alt="image" src="https://github.com/decompme/decomp.me/assets/737603/509c930e-2b0c-43b1-bb61-5529ca4a570c">

Constrast values: (counting everything below 3.0 as a failure, though this should probably be more like 4.5)
<img width="312" alt="image" src="https://github.com/decompme/decomp.me/assets/737603/f5c81e94-48e5-4b90-904d-ad170a51aa32">

mkst:
Is this a duplicate of #https://github.com/decompme/decomp.me/issues/678 ?

PR #773, opened 2023-05-27
Diff text color overhaul

bismurphy:
Diff colors can be hard to read on some backgrounds, especially the existing light cyan. This changes colors from an arbitrary-seeming set to a programmatic set, generated through equally spaced HSV values. This should approximate a strong set of distinguishable colors. These colors are easy to read on both light and dark modes, as shown in this test screenshot:

https://cdn.discordapp.com/attachments/897075447534354462/1112075654947672134/image.png

### DIFF
--- a/frontend/src/components/Diff/Diff.module.scss
+++ b/frontend/src/components/Diff/Diff.module.scss
@@ -123,32 +123,7 @@
 
 .source_other { font-style: italic; }
 .source_line_num { font-style: italic; }
- color:#cd5252
-}
-.Diff_rotation1__cQ_cP {
- color:#cda452
-}
-.Diff_rotation2__KBLKI {
- color:#a4cd52
-}
-.Diff_rotation3__35Yxv {
- color:#52cd52
-}
-.Diff_rotation4__G2xR5 {
- color:#52cda4
-}
-.Diff_rotation5__kkcNj {
- color:#52a4cd
-}
-.Diff_rotation6__5x_yp {
- color:#5252cd
-}
-.Diff_rotation7__x8lnZ {
- color:#a452cd
-}
-.Diff_rotation8__TYVPj {
- color:#cd52a4
-}
+
 .rotation0 { color: rgb(205,82,82); }
 .rotation1 { color: rgb(205,164,82); }
 .rotation2 { color: rgb(164,205,82); }
--- a/frontend/src/components/Diff/Diff.module.scss
+++ b/frontend/src/components/Diff/Diff.module.scss
@@ -124,15 +124,15 @@
 .source_other { font-style: italic; }
 .source_line_num { font-style: italic; }
 
-.rotation0 { color: rgb(205,82,82); }
-.rotation1 { color: rgb(205,164,82); }
-.rotation2 { color: rgb(164,205,82); }
-.rotation3 { color: rgb(82,205,82); }
-.rotation4 { color: rgb(82,205,164); }
-.rotation5 { color: rgb(82,164,205); }
-.rotation6 { color: rgb(82,82,205); }
-.rotation7 { color: rgb(164,82,205); }
-.rotation8 { color: rgb(205,82,164); }
+.rotation0 { color: rgb(205, 82, 82); }
+.rotation1 { color: rgb(205, 164, 82); }
+.rotation2 { color: rgb(164, 205, 82); }
+.rotation3 { color: rgb(82, 205, 82); }
+.rotation4 { color: rgb(82, 205, 164); }
+.rotation5 { color: rgb(82, 164, 205); }
+.rotation6 { color: rgb(82, 82, 205); }
+.rotation7 { color: rgb(164, 82, 205); }
+.rotation8 { color: rgb(205, 82, 164); }
 
 .highlighted {
     color: white;
--- a/frontend/src/components/Diff/Diff.module.scss
+++ b/frontend/src/components/Diff/Diff.module.scss
@@ -123,15 +123,41 @@
 
 .source_other { font-style: italic; }
 .source_line_num { font-style: italic; }
-.rotation0 { color: magenta; }
-.rotation1 { color: cyan; }
-.rotation2 { color: rgb(0, 212, 0); }
-.rotation3 { color: red; }
-.rotation4 { color: rgb(103, 106, 255); }
-.rotation5 { color: lightpink; }
-.rotation6 { color: blue; }
-.rotation7 { color: lightgreen; }
-.rotation8 { color: grey; }
+ color:#cd5252
+}
+.Diff_rotation1__cQ_cP {
+ color:#cda452
+}
+.Diff_rotation2__KBLKI {
+ color:#a4cd52
+}
+.Diff_rotation3__35Yxv {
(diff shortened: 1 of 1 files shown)


ethteck:
image
 before

ethteck:
image

after

frontend/src/components/Diff/Diff.module.scss
```diff
@@ -123,15 +123,16 @@
...
+
+.rotation0 { color: rgb(205,82,82); }
+.rotation1 { color: rgb(205,164,82); }
+.rotation2 { color: rgb(164,205,82); }
+.rotation3 { color: rgb(82,205,82); }
+.rotation4 { color: rgb(82,205,164); }
+.rotation5 { color: rgb(82,164,205); }
+.rotation6 { color: rgb(82,82,205); }
+.rotation7 { color: rgb(164,82,205); }
+.rotation8 { color: rgb(205,82,164); }
```
nanaian, line 135: Good work, thank for the contribution! Can you add spaces after the commas, i.e. `rgb(205, 82, 82)`?

nanaian (changes requested):

frontend/src/components/Diff/Diff.module.scss
bismurphy: Done! New commit does this.

mkst:
LGTM?

marijnvdwerf:
First of all, thanks for your effort. When I created #757, I wasn't expecting anyone to actually work on it.

Despite that, I disagree with the proposed changes. While they do increase the minimal contrast ratio for light mode, most colours still don't meet a minimum constrast ratio of 3.0. In addition, the changes have a negative effect on the dark mode contrast ratios.

image

I see that the colours have all been defined in `hsl`, with the same lightness. This lightness however doensn't match with how people perceive colours (so yellow and blue are the same lightness, but vary wildly in visual brightness). If you want colours to appear similarly bright, you should look into something like `LCh` instead.

Additionally, I consider the hue-based sorting of colours a downgrade. The diff colours become only really useful when there's just a few of them, so you want e.g. 4 colours to still cover most of the colour wheel. The IBM Carbon page describes that for their datavis palette.

> The colors of this palette should be applied in sequence strictly as described below. *The sequence is carefully curated to maximize contrast between neighboring colors to help with visual differentiation.*
> https://carbondesignsystem.com/data-visualization/color-palettes/#categorical-palettes

In conclusion, I think we need to:
- Use separate colour schemes for light and dark mode
- Ensure the colours contrast with the background
- Make sure that the colours differ enough with each other.
- Try to take into account colour-blindness. (so don't use red and green with the same intensity etc.)

I think most of this is best done by using an existing colour scheme. I did look into both IBM Carbon and Adobe Spectrum (since they both have datavisualisation palettes), but I wasn't terribly excited about either of them.

image
image

mkst:
Are you using a tool or a website to generate these images? I'm keen to curate separate light/dark themes with sufficient contrast but feels like i'm missing a tool to check multiple colours at once

nanaian:
> Use separate colour schemes for light and dark mode

Prefix `.dark` to select for dark theme users only

marijnvdwerf:
@mkst I threw something together locally. Pushed it to https://github.com/marijnvdwerf/playground-colour-contrast now.

I think for the diff we should consider to switch to using a background colour (possible with a profile setting to switch to blue/yellow or blue/red), to differentiate a bit between "major diffs" and regalloc diffs.

Since contrast on light-grey seems super hard, I think we should make sure that in light mode, all editors have a `#fff` background. The should hopefully help a _little_ bit.

ethteck:
Is that something you wanna do in this PR @marijnvdwerf ? I've kinda lost track of what this PR needs (assuming we still plan to take it)

marijnvdwerf:
I think if we only use these colours for light mode, it's still an improvement over the current situation.

The "ideal" solution would require some more thought and testing (Decide on a library starting point, maybe get some automated testing in, get some visually-impaired users involved)

Don't know how @nanaian wants to handle light/dark, and which one should be considered an override. I'd be okay with something like this. I'd like to see the new colours defined as hsl, so the logic behind it is more visible.

```
.rotation0 { color: magenta; }
.rotation1 { color: cyan; }
.rotation2 { color: rgb(0, 212, 0); }
.rotation3 { color: red; }
.rotation4 { color: rgb(103, 106, 255); }
.rotation5 { color: lightpink; }
.rotation6 { color: lightcyan; }
.rotation7 { color: lightgreen; }
.rotation8 { color: grey; }

.light {
    .rotation0 { color: rgb(205, 82, 82); }
    .rotation1 { color: rgb(205, 164, 82); }
    .rotation2 { color: rgb(164, 205, 82); }
    .rotation3 { color: rgb(82, 205, 82); }
    .rotation4 { color: rgb(82, 205, 164); }
    .rotation5 { color: rgb(82, 164, 205); }
    .rotation6 { color: rgb(82, 82, 205); }
    .rotation7 { color: rgb(164, 82, 205); }
    .rotation8 { color: rgb(205, 82, 164); }
}
```

mkst:
I feel like it might be good to take a step back, catalogue all colours used in the diff (because there are more than just those rotation ones), and then come up with 2 schemes (1 light, 1 dark) that have appropriate contrast against the background (and each other) and *then* raise a PR to implement the changes. 

Or, we just merge this and iterate :)

bismurphy:
> I feel like it might be good to take a step back, catalogue all colours used in the diff (because there are more than just those rotation ones), and then come up with 2 schemes (1 light, 1 dark) that have appropriate contrast against the background (and each other) and _then_ raise a PR to implement the changes.
> 
> Or, we just merge this and iterate :)

I would be inclined toward merge and iterate. Merging means the half-solution will be "live", so people will be affected by it and more likely to contribute to improve their own experience. Something sitting in analysis limbo is less likely to have buy-in for anyone to put time and energy into refining.

ethteck:
> > I feel like it might be good to take a step back, catalogue all colours used in the diff (because there are more than just those rotation ones), and then come up with 2 schemes (1 light, 1 dark) that have appropriate contrast against the background (and each other) and _then_ raise a PR to implement the changes.
> > Or, we just merge this and iterate :)
> 
> I would be inclined toward merge and iterate. Merging means the half-solution will be "live", so people will be affected by it and more likely to contribute to improve their own experience. Something sitting in analysis limbo is less likely to have buy-in for anyone to put time and energy into refining.

Could we make it so this PR just changes the colors for light mode? (or is that the current situation? I've lost track lol)

nanaian:
> Don't know how @nanaian wants to handle light/dark, and which one should be considered an override. I'd be okay with something like this

LGTM

nanaian:
We can merge and iterate if its OK on both light & dark but we need to revisit this

ethteck:
I think hamster was implying the new colors are worse for the dark theme, which is why I think we should have this PR only affect light mode if we're gonna merge it

ethteck:
I think for now we should continue discussing the best ultimate approach to this issue, but since the very imminent problem of light mode colors being bad has been fixed in #797 , I'm going to close this for now. Thanks for all your effort @bismurphy as well as everyone's useful input

Issue #782, opened 2023-06-10
`JSON.parse` runtime exception


ribbanya:
I keep encountering the following error:
![image](https://github.com/decompme/decomp.me/assets/5190964/d3b26bde-51a2-43d7-8730-c2fa9c24cc4c)

Seems to occur on the entire website, but only on my main Firefox profile. A Private Browsing window doesn't experience it.

I have one cookie that's 16.4 MB but I don't want to leak anything. Let me know what would be helpful.

```
User-Agent: Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/113.0
```

mkst:
Does it happen on a particular scratch? Is it reproducible or intermittent? Is there anything interesting in the developer console (hit F12, select console).

This usually happens when the backend returns something that isn't JSON (i.e it had an error), but interesting if you're experiencing different behaviour in private browsing. 

Maybe related to local storage (i.e. bad data in a scratch context?). Would be good to get to the bottom of this rather than just asking you to clear site cookies/data.

ribbanya:
It seems to occur on every scratch, and the homepage, etc. Initially the page loads and then is replaced with the exception.

Turns out there's quite a bit going on in the console.
```
GET https://decomp.me/api/user 1185-1963ad663e0d73ef.js:1:9221
GET https://decomp.me/api/stats 1185-1963ad663e0d73ef.js:1:9221
GET https://decomp.me/api/user/scratches?page_size=16 1185-1963ad663e0d73ef.js:1:9221
GET https://decomp.me/api/scratch?page_size=30 1185-1963ad663e0d73ef.js:1:9221
TypeError: can't convert undefined to object .excmds_content.generated.ts:2097:27
SyntaxError: JSON.parse: unexpected character at line 2 column 1 of the JSON data 8944-61dca2a8f1727bdc.js:1:4476
    NextJS 87
Error boundary caught an error: SyntaxError: JSON.parse: unexpected character at line 2 column 1 of the JSON data 
[...]

PR #783, opened 2023-06-12
Fix server error upon processing user with revoked github token

ConorBobbleHat:
Fixes the root cause behind #782.

Right now, `GitHubUser.details` uses the user's authentication token to fetch their own details. In doing so, it doesn't check if the token's been revoked by its owner, and errors out if it has.

As all the details we're looking for are publicly available, this PR switches to fetching the user's details sans an authentication token. Insofar as I can tell, this is the only such place we're using an auth token unnecessarily.

This is a little bit of a band-aid fix because it ignores the fact there are still other ways fetching a user's information can fail (although they're rarer still). Happy for greater error checking to be included in this PR, or to leave it as a future issue.

### DIFF
--- a/backend/coreapp/models/github.py
+++ b/backend/coreapp/models/github.py
@@ -65,7 +65,7 @@ class GitHubUser(models.Model):
         if cached:
             return cached
 
-        details = Github(self.access_token).get_user_by_id(self.github_id)
+        details = Github().get_user_by_id(self.github_id)
 
         cache.set(cache_key, details, API_CACHE_TIMEOUT)
         return details


nanaian:
> As all the details we're looking for are publicly available

Is this certainly the case for things like email? Or does the response differ a bit
I don't think we use it but just so we know for future

ConorBobbleHat:
From https://docs.github.com/en/rest/users/users?apiVersion=2022-11-28#get-a-user, the `/users/<username>` endpoint
> Provides publicly available information about someone with a GitHub account.

The waters are muddied slightly by the fact we're using an undocumented version of the endpoint (`/user/<user id>`) instead, but - at least as I write this - they resolve to exactly the same response.

ConorBobbleHat:
in the process of looking that up, I did stumble across this: https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#rate-limiting

> For unauthenticated requests, the rate limit allows for up to 60 requests per hour. Unauthenticated requests are associated with the originating IP address, and not the person making requests.

That would be down from 15,000 an hour for a user-authenticated request. This one may need to back to the drawing board for a bit.

mkst (approved):
It's a rare scenario, but it's already happened once - so I say lgtm :)

nanaian (approved):
Looks good. Might be worth changing `__str__` to not call `details` given that its more liable to fail with a timeout exception now. It was stupid of me to have `__str__` possibly make a network request in the first place tbh. https://github.com/decompme/decomp.me/blob/4106e8c0e5cc2fcbb16cbd5a8400c77318f94dc3/backend/coreapp/models/github.py#L95-L96

ethteck:
I'm cool with merging this after that last comment is addressed

nanaian:
I think itd be fine to just report the user id

backend/coreapp/models/github.py
```diff
@@ -65,13 +66,31 @@ def details(self) -> NamedUser:
...
+            # This *does* have a much stricter rate limit of 60 requests per minute,
+            # which is why we tried using an auth token first
+            details = Github().get_user_by_id(self.github_id)
 
         cache.set(cache_key, details, API_CACHE_TIMEOUT)
         return details
 
     def __str__(self) -> str:
-        return "@" + self.details().login
+        return "@" + self.user.username
```
ConorBobbleHat, line 93: Happy to change this over to user ID if I'm missing something, but: I think this should preserve the behavior we have now, but without the network request?

ethteck: seems good to me!

nanaian:
Good to merge once tests pass ✅

ethteck:
The 8 tests that failed are the new compilers which will be fixed in #800 . merging

PR #784, opened 2023-06-12
Add ARMCC 4.1 Build 791

fruityloops1:
(the contents of the download link were updated)


ethteck:
@fruityloops1 this one seems to be failing - if you look at the full test and build logs, you can see the error info

fruityloops1:
i updated the download again, it should work now if you redo the checks

PR #787, opened 2023-06-18
Add ee-gcc2.9-991111-01 compiler

entriphy:
This compiler is required to get this scratch this to match: https://decomp.me/scratch/oPCWT

I also added a few notes at the top of that scratch detailing the differences between 991111, 991111-01, and the 991111 Windows build that the site currently uses (I'll call this 991111-win). 991111 can be compiled from source so Wine wouldn't have to be used, though it produces different code than 991111-win, so I'm not sure how to proceed with adding that compiler version.

(technically 991111 should be `ee-gcc2.9-991111` and 991111-win should be moved to a separate compiler ID like `ee-gcc2.9-991111-win` to avoid confusion since 991111-win seems to be a compiler version in between 991111 and 991111-01, though that means certain scratches that need 991111-win would have to change the compiler used, i.e. https://decomp.me/scratch/QIQfY)

### DIFF
--- a/backend/compilers/download.py
+++ b/backend/compilers/download.py
@@ -636,6 +636,7 @@ def download_ps2():
 
     ps2_compilers = {
         "ee-gcc2.9-990721": "https://cdn.discordapp.com/attachments/1067192766918037536/1067306679806464060/ee-gcc2.9-990721.tar.xz",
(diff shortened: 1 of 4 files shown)


ethteck:
If you think the id should be renamed, which you make a good case for, would you mind making a db migration that does that? you can find an example in one of the previous migrations

entriphy:
Should be good now. I was also able to compile a Linux build of 991111-win, so I just called it `ee-gcc2.9-991111a`.

Issue #468, opened 2022-06-06
Support for MS-DOS and Open Watcom


malespiaut:
Hello,

I would like to request support for MS-DOS using the Open Watcom compiler.

Best regards.

ethteck:
Hello. Unfortunately, we're kinda a long way from supporting that platform because our asm-differ tool only supports certain architectures. Not to say it could never happen, but you'd need to first add support there.

Do you know of any decomp projects using this compiler / platform? I'd be curious what their workflow is and what kinds of tools they're using.

malespiaut:
Hello,

Thank you for your fast and detailed answer.

I thought that the Godbolt decompiler was supporting Open Watcom, but upon inspection, this isn't the case.

The project https://github.com/dethrace-labs/dethrace is the only one I know that makes use of the Open Watcom compiler, but I can't say for sure that it's purely a decompilation project. It would be best to ask @dethrace-labs directly.

I'll keep an eye on that issue, in the meantime.

OmniBlade:
Not public, but I'm aware of a project to binary match the original Command and Conquer windows binaries that used watcom 10.5. Workflow is using custom scripts to dump asm from IDA. Its helped a lot by the fact so much of the original code was released open source as part of the Command and Conquer remaster release, but there are lots of undocumented changes in the code from the original and code for Sole Survivor was not released.

ethteck:
Done in #794

PR #794, opened 2023-06-27
Add MS-DOS platform and Watcom compilers.

OmniBlade:
Provisional implementation, need some clarification on where the tools should actually go and if they should be pulled down in downloads.py as I currently have it or in the Dockerfile for the backend. Also, the MS-DOS icon isn't scaling right for me and I can't figure out how to fix it.


ethteck:
> if they should be pulled down in downloads.py

They indeed should, and it looks like you've set it up that way.

Can you explain what you mean by the logo not scaling right? I'll try to test locally soon. Also, be sure to see the black failures / run black locally if you want it to auto-fix the formatting issues.

ethteck:
there are a bunch of (all?) compilers failing in the non-docker test, and 3 failing in the docker test. I'm not even sure if all 3 are new compilers either. very weird

backend/Dockerfile
```diff
@@ -101,11 +103,12 @@ ENV ENABLE_NDS_SUPPORT=${ENABLE_NDS_SUPPORT}
 ENV ENABLE_PS1_SUPPORT=${ENABLE_PS1_SUPPORT}
 ENV ENABLE_SWITCH_SUPPORT=${ENABLE_SWITCH_SUPPORT}
 ENV ENABLE_SATURN_SUPPORT=${ENABLE_SATURN_SUPPORT}
+ENV ENABLE_MSDOS_SUPPORT=${ENABLE_MSDOS_SUPPORT}
 
 RUN python3.10 -m pip install requests tqdm \
     && python3.10 /compilers/download.py \
     && rm -rf /compilers/download_cache/
-
+
```
mkst, line 111: Could you trim the trailing whitespace in this file?

backend/coreapp/compilers.py
```diff
@@ -893,6 +898,57 @@ def preset_from_name(name: str) -> Optional[Preset]:
     cc=MWCCARM_CC,
 )
 
+WATCOM_105_C = WatcomCompiler(
+    id="wcc10.5",
+    platform=MSDOS,
+    cc='${WINE} "${COMPILER_DIR}/binnt/wcc386.exe" -zq -i="..\\compilers\\wcc10.5\\h" -i="..\\compilers\\wcc10.5\\h\\nt" ${COMPILER_FLAGS} -fo"object.o" "code.c"',
```
mkst, line 904: Can you replace the `"object.o"` with `"${OUTPUT}"` and `"code.c"` with `"${INPUT}"` to avoid the hardcoding?

backend/coreapp/platforms.py
```diff
@@ -722,6 +736,7 @@ def from_id(platform_id: str) -> Platform:
         "gba": GBA,
         "macos9": MACOS9,
         "macosx": MACOSX,
-        "n3ds": N3DS,
+        "msdos" : MSDOS,
+        "n3ds": N3DS
```
mkst, line 740: can you add the trailing `,` back in?

backend/coreapp/compilers.py
OmniBlade: I tried with input/output variables but they are passed as a Unix path and watcom treats '/' as an optional way of indicating compiler options leading to errors of unknown options. I didn't see any obvious examples where this was worked around in the existing compilers but I'm open to suggestions on how to handle it better as I agree hard coding it isn't ideal.

mkst:
This PR closes https://github.com/decompme/decomp.me/issues/468

backend/coreapp/compilers.py
mkst: Ah fair enough, that's frustrating that wine doesn't automagically fix up your paths.

OmniBlade:
> > if they should be pulled down in downloads.py
> 
> They indeed should, and it looks like you've set it up that way.
> 
> Can you explain what you mean by the logo not scaling right? I'll try to test locally soon.
> 
> Also, be sure to see the black failures / run black locally if you want it to auto-fix the formatting issues.

I think I've addressed the formatting and the comments relating to hard coded paths. I would like someone to verify the icon works correctly for them though before I take this out of draft. The issue I see is that it only displays the top left corner of the icon like its really zoomed in, but I thought I'd fixed that by adjusting the scaling in inkscape.

mkst:
The logo looks OK to me:
image

ethteck (approved):

ethteck:
@OmniBlade lgtm, but black has a couple more issues to resolve. Also, if you could refrain from force-pushing, it helps reviewers keep track of new changes over time

ethteck:
weird..other simple tests failed too in the most recent CI build. I'll look into it after the black things are fixed

ethteck:
@OmniBlade I think the same issues still remain, or at least the same lines

PR #801, opened 2023-07-04
Use doldecomp/melee as source for frank.py

r-burns:
Melee is usually the main motivator for frank updates, as the vast majority of melee code needs 1.2.5e, so most frank updates hit melee first. I'm also currently held up on syncing pikmin's frank.py with melee's, as the latest update to frank needs python 3.8 and pikmin's CI is on 3.7. https://github.com/projectPiki/pikmin/pull/9


EpochFlame:
yeah it’s probably for the best to use melee’s

PR #802, opened 2023-07-06
Implement Win 9x platform and msvc compilers.

OmniBlade:
Initial punt at adding MSVC compilers, adds all varieties of MSVC 6 I could find which covers a wide range of retro games and software for Windows 9x upwards as well as MSVC 7. The compilers work, but there is an issue with read only wine temp directories which I don't think I've solved in the most appropriate wa as some info can leak out of the sandbox in the current config. Also, I've had exceptions with the differ on some mangled names.


ItzSwirlz:
MSVC 4.20 would be good for LEGO Island. There's a portable version here: https://github.com/itsmattkc/msvc420

ethteck:
Let's add MSVC 4.20 & 7.1 in a follow-up PR, but yeah I am eager to get these in shoon

PR #804, opened 2023-07-09
Add a way to manually mark a scratch as matching (fixes #401)

### LINKED ISSUE #401: Manual "match" override
Allow project members and site admins to manually mark a scratch as matching, awarding #385 for the match

ethteck:
Fixes #401


nanaian (approved):
Still skeptical of the usecase but cod seems fine

ethteck:
@nanaian just in case you care to see, here's an example where this will help: https://decomp.me/scratch/hGfAk

There are several occasions where, due to target asm being wrong or due to known issues with the diffing / disassembly, it's helpful to still be able to say "this is matching, even if the site doesn't think it is"

PR #807, opened 2023-07-15
Add 1.2.5n compiler, and use it for melee/pikmin

r-burns:


ethteck:
Everything in here looks good! I was thinking it might be nice to make a migration to the new compiler for all scratches that are on the old one

PR #834, opened 2023-09-03
Add msvc4.0/4.2 support + lego island preset

ItzSwirlz:
Using MattKC's repos. They are named differently and dont have a release tagged so the dl is a bit different. Also added LEGO Island's preset from the CMakeLists (https://github.com/isledecomp/isle/blob/master/CMakeLists.txt)


backend/compilers/download.py
```diff
@@ -975,6 +975,23 @@ def download_win9x():
...
         download_tar(url=url, dest_name=compiler)
 
+    # For the repo these compilers are stored in they need a few location adjustments for neatness, and permissions set to executable
+    download_zip(
+        url="https://github.com/itsmattkc/MSVC400/archive/refs/heads/master.zip",
+        dest_name="msvc40",
+    )
+    shutil.move(COMPILERS_DIR / "MSVC400-master", COMPILERS_DIR / "msvc4.0")
+    shutil.move(COMPILERS_DIR / "msvc4.0/BIN", COMPILERS_DIR / "msvc4.0/Bin")
+    os.system("chmod -R +x " + str(COMPILERS_DIR / "msvc4.0" / "Bin"))
```
ethteck, line 985: Could you call set_x on the necessary files within Bin instead?

backend/compilers/download.py
```diff
@@ -975,6 +975,23 @@ def download_win9x():
...
+
+    download_zip(
+        url="https://github.com/itsmattkc/MSVC420/archive/refs/heads/master.zip",
+        dest_name="msvc42",
+    )
+    shutil.move(COMPILERS_DIR / "MSVC420-master", COMPILERS_DIR / "msvc4.2")
+    shutil.move(COMPILERS_DIR / "msvc4.2/bin", COMPILERS_DIR / "msvc4.2/Bin")
+    os.system("chmod -R +x " + str(COMPILERS_DIR / "msvc4.2" / "Bin"))
```
ethteck, line 993: here2

ethteck (changes requested):

PR #843, opened 2023-09-09
Support libraries in decomp.me

### LINKED ISSUE #842: Make some common libraries available to the compiler
It'd be very useful if decompme could make some libraries available, such that a scratch could depend on, say, DirectX 8 and use its `#include` in the context. This would make decompme easier to use (as the context could then `#include` the necessary headers easily, limiting the need for `makectx` scripts), and reduce total disk space used by scratches (due to not duplicating a bunch of definitions from common libraries).

# Rough design

For the purpose of this feature, the header of libraries need to be made available to the compilers. We only care about the headers since decompme doesn't do any linking, it merely operates on (unlinked) object files. The libraires would be stored in a new `libraries` directory, next to the `compilers`. Within it, the headers would be stored in `<name>/<version>/include` folders. (The `/include` being here to support future extensions if we ever want to support linking). So for instance, we could have the following structure

```
.
└── libraries
    ├── directx
    │   ├── 8.0
    │   │   └── include
    │   │       └── dx3d.h
    │   ├── 8.0a
    │   │   └── include
    │   │       └── dx3d.h
    │   ├── 8.1
    │   │   └── include
    │   │       └── dx3d.h
    │   ├── 9.0
    │   │   └── include
    │   │       └── dx3d.h
    │   └── 9.0-summer2003
    │       └── include
    │           └── dx3d.h
    └── sdl
        └── 2.0.0
            └── include
                └── SDL.h
```

A new script, `backend/libraries/download.py`, would be responsible for acquiring and storing those libraries on the local filesystem. This script would likely share some code with the existing backend/compilers/download.py, as they largely share thee same task.

A new API route would need to be made available to expose the supported libraries. This API would iterate on known libraries, check if they were downloaded, and return them as JSON for the frontend to consume, similar to the existing `compilers` endpoint implemented [here](https://github.com/decompme/decomp.me/blob/a7608e80860580a155d6806b2c8ea95af04e2a77/backend/coreapp/views/compilers.py#L17). 

Another API route would need to be added to add and delete a library to/from a scratch. The backend should then add/remove it to a (new) database row of the scratch that lists all the currently used libraries/versions. When compiling a scratch, the backend would get that list of libraries/versions, and pass the necessary flags to the compiler to let it find the necessary headers (for instance, it'd pass `-I` to GCC).

On the frontend side, we'll need a way to add libraries to the current scratch. This could be done in various ways, but the easiest would probably be to add it to the `Option` tab. We could show a list of currently used libraries (with a "delete" button next to each), along with a combobox to select a library/version to add.

# Unresolved questions / Alternative designs

- Should we "limit" libraries to a particular platform/compiler? Some libraries only make sense to use with certain platforms/compilers (for instance, directx only makes sense for win9x compilers, and make no sense for other platforms).
- Instead of adding a new field to the scratch to store the list of libraries, we could just add the necessary flags directly to `compiler_flags`, requiring less changes to the overall codebase.

roblabla:
This MR introduces libraries in decomp.me. This is a new section under "Options" that allow adding libraries to the scratch, which allows the user to `#include` the headers provided by that lib.

Fixes #842

Example:

image

This is ready for review.


roblabla:
There's one mypy error that I just can't figure out for the life of me. If someone can find a fix, I'd be grateful.

Otherwise, this feature is ready for review.

backend/compilers/download.py
```diff
@@ -237,6 +238,133 @@ def get_compiler(
     return True
 
 
+def get_library(
```
mkst, line 241: I feel like the library downloading should be a separate script from the compilers downloading. Dunno what other people's opinions are though.

ethteck: ah, yeah that didn't really register to me, but I agree

roblabla: Yeah that's fair. It used to reuse a lot of the infrastructure from compilers/download.py so it made sense to have a single script, but now that it does its own stuff, it can move to its own script.

roblabla: Fixed. I also updated the docs to explain what the new libraries folder does.

nanaian (approved):
Frontend changes lgtm

ethteck:
image
I think the library flag(s) that get appended should always start with a space

roblabla:
> I think the library flag(s) that get appended should always start with a space

Nice catch, I added a space, and also moved the libraries to be within `sandbox.quote_options`

PR #860, opened 2023-09-18
API - Allow filtering scratches based on preset

### LINKED ISSUE #859: Expose compiler preset to the search API endpoint
I made a simple python script that uses the `https://decomp.me/api/scratch` endpoint to generate a simple HTML page that tells me which DKR functions are on decomp.me or not: https://davidsm64.github.io/snapshot.html

However, a minor issue that popped up is that there are a few functions that have the same name as functions from other N64 games. Currently I'm using the `compiler`, `language`, and `platform` to determine if a function is a DKR function. 

```python
def is_valid_dkr_function(result):
    return result['compiler'] == 'ido5.3' and result['language'] == 'C' and result['platform'] == 'n64'
```

I'd like the results to also include the compiler preset name, so that I can discard the false positives by checking if the compiler preset is set to the DKR preset.

mkst:
`filter`ing is broken at the moment (see https://stackoverflow.com/a/73887435/7290888).

This doesnt currently work:
```
curl "https://decomp.me/api/scratch?compiler=ido5.3" | jq
```

This PR adds the ability to filter scratches based on the preset (also adds preset to the terse serializer)
```
curl "http://localhost:8000/api/scratch?preset=AeroGauge"
# i.e.
curl "https://decomp.me/api/scratch?preset=AeroGauge" | jq
```

Closes #859.

### DIFF
--- a/backend/coreapp/serializers.py
+++ b/backend/coreapp/serializers.py
@@ -242,6 +242,7 @@ class TerseScratchSerializer(ScratchSerializer):
             "project",
             "project_function",
             "parent",
+            "preset",
         ]
 
 
--- a/backend/coreapp/views/scratch.py
+++ b/backend/coreapp/views/scratch.py
@@ -294,7 +294,7 @@ class ScratchViewSet(
 ):
     queryset = Scratch.objects.all()
     pagination_class = ScratchPagination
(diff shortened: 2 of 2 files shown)


DavidSM64:
Thank you!

PR #868, opened 2023-09-24
Add "Metal Gear Solid (overlays)" preset

nocato:
Add a new "Metal Gear Solid (overlays)" preset that uses `-G0` flag instead of `-G8` flag, as was the case in the "Metal Gear Solid" preset  (preset for the main executable). 

Additionally add `-Wall` flag, because we use it very often in our scratches.

### DIFF
--- a/backend/coreapp/compilers.py
+++ b/backend/coreapp/compilers.py
@@ -1384,6 +1384,11 @@ _all_presets = [
         "-O2 -G8",
     ),
     Preset(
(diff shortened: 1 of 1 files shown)


ethteck:
I have a maybe strange request:  could you make a new preset and add overlay or something to the name? I'd like to be able to pair presets to the old scratches and have them still match, once the projects feature is more fleshed out

nocato:
@ethteck Sure, modified the PR to add a new preset instead.

PR #869, opened 2023-09-27
Add presets for The Wind Waker

MrPolymorph:


ethteck:
Black failed here - if you click the failure, you should see what needs to be fixed

PR #874, opened 2023-10-01
No podman in compilers download

roblabla:
Podman/Docker don't work well in docker-on-macos for whatever reason, leading to being unable to download the compilers.

Downloading the images manually is easy enough to do, the OCI Distribution specification is easy enough to follow (see [this](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pull).


mkst:
This is great! I think you can also remove the docker/podman requirements from the poetry file :+1:

backend/compilers/download.py
```diff
@@ -154,97 +99,96 @@ def get_compiler(
...
+    mime = manifest['mediaType']
+    headers['Accept'] = mime
+    url = f"https://{docker_registry}/v2/{github_repo}/{docker_image}/manifests/{digest}"
+    resp = requests.get(url, headers=headers, timeout=10)
+    if resp.status_code != 200:
+        return None
+    data = resp.json()
+
+    # Then, find the layer, and download it.
+    layer = data['layers'][0]
```
mkst, line 142: does this assume that the compiler image is only going to have a single layer (currently true, but not sure if it will always be the case)?

roblabla: Yes. I could adjust this to extract the layers one after the other though.

roblabla: Implemented this.

roblabla:
> This is great! I think you can also remove the docker/podman requirements from the poetry file 👍

Done. The first commit now bumps poetry to 1.6.1 (without any other changes), while the second commit removes the docker/podman dependencies (along with implementing the raw download)

nanaian (approved):
I get these warnings- but it works!
```
$ poetry run python compilers/download.py
Unable to get image manifest for n64/ido7.1:latest from decompme/compilers: {"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest unknown"}]}

Unable to get image manifest for n64/ido5.3:latest from decompme/compilers: {"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest unknown"}]}

Unable to get image manifest for n64/gcc2.7.2kmc:latest from decompme/compilers: {"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest unknown"}]}

Unable to get image manifest for n64/gcc2.8.1pm:latest from decompme/compilers: {"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest unknown"}]}

Unable to get image manifest for switch/clang-4.0.1:latest from decompme/compilers: {"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest unknown"}]}

Unable to get image manifest for switch/clang-8.0.0:latest from decompme/compilers: {"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest unknown"}]}

2023-10-02 21:25:46.677 INFO main Updated 6 / 6 compiler(s) in 79.09 second(s)
```

mkst:
Yeh I think that's ok, these particular images have linux/darwin flavours, but by default we lookup the "non-arch name" image

roblabla:
I'll try to silence the error if we have an os-specific flavor for the image, as I feel it may be a bit misleading.

So left on my TODO:

- Silence the warning
- Figure out what the fudge is going on in CI (see https://github.com/decompme/decomp.me/actions/runs/6372041695/job/17294505786?pr=874 - I'll probably need to ssh into the runner to figure this out) 

I'll try to do that tomorrow.

mkst:
I'll suggest a change when I'm at my computer, or if you look at what I did for the podman pr, just undo the change:)

backend/docker_entrypoint.sh
```diff
@@ -10,13 +10,9 @@ poetry config virtualenvs.path /backend/virtualenvs
 
 poetry install
 
-podman system service --time=0 unix:///tmp/podman.sock &
-
-poetry run /backend/compilers/download.py --podman
+poetry run /backend/compilers/download.py
 poetry run /backend/libraries/download.py
```
mkst, line 14: could we also do
```
poetry run libraries/download.py
```
here, as it looks like it was missed from the libraries PR?

PR #875, opened 2023-10-02
upgrade frontend dependencies

### LINKED ISSUE #763: Internal links sometimes just don't work
From the credits page, to user pages, for example

nanaian:
Upgrades all* frontend dependencies to their latest versions. I believe this fixes #763.

*except radix-ui/colors, because they changed from hsl to hex codes & that would break how we derive colors

nb there is a deprecation in stylelint (`yarn lint`) that recommends we start using Prettier. I agree, but that's out of scope for this PR


ethteck:
Clicking "reset anonymous appearance" on the settings page causes some loading, one name to popup, and then another. is there a reason it's happening twice? https://frontend-git-upgrade-frontend-decompme.vercel.app/settings/account

ethteck:
nm, it's unrelated to this. it's just something weird the vercel preview is doing

PR #879, opened 2023-10-06
Remove library suspense...?

ethteck:


ethteck:
this seems to be working, based on the preview deployment. is it really this simple?

roblabla:
> is it really this simple?

Looks like it. The fix LGTM.

nanaian:
This is cheating, as we should show a loading state or something when `!data`. As a hotfix its fine

roblabla:
> This is cheating, as we should show a loading state or something when !data. As a hotfix its fine

I guess something that could be done is have `useLibraries` return null, and if it does, make the `Select` read-only with a single "Loading" value, or something along those lines

ethteck:
Returning an empty array means we just display the "--" entry if none are loaded. I wasn't able to observe any state where the stuff wasn't loaded. so I'm gonna merge this, but we can enhance it later if we notice any weirdness

PR #881, opened 2023-10-07
Refactor libraries/download.py

roblabla:
This PR:

- avoids the warning about git init without naming the branch when downloading libs:
```
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: 
hint:     git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
hint:     git branch -m <name>
```

- Removes an unused `--libraries` argument
- Adds an error if the downloaded repository does not contain an `include` directory

Along with some misc cleanup of the script.

### DIFF
--- a/backend/libraries/download.py
+++ b/backend/libraries/download.py
@@ -37,7 +37,7 @@ def get_library(
     if "git" in download_info:
         # Download with git. Get url and ref, and download using git clone.
         url = download_info["git"]["url"]
-        branch = download_info["git"]["branch"]
+        branch = str(download_info["git"]["branch"])
         logger.debug("Using git to download library at %s branch %s", url, branch)
 
         # Recreate repository if force is set.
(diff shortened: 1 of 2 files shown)


mkst:
Not sure why the CI failed, recording it here for reference:
```
======================================================================
FAIL: test_family_etag (coreapp.tests.ScratchDetailTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/decomp.me/decomp.me/backend/coreapp/tests.py", line 1090, in test_family_etag
    self.assertNotEqual(etag, response.headers.get("Etag"))
AssertionError: '"-1200886303259210849"' == '"-1200886303259210849"'
```

PR #897, opened 2023-10-24
Presets in the DB

ethteck:
Added:
* new Preset model and a migration to move all previously hard-coded presets into the db
  * This includes some new fields that currently aren't used (decompiler_flags, assembler_flags) and renames `flags` to `compiler_flags`
  * Scratches can now be created with just a preset id, and this populates all the relevant scratch settings from the DB
  * Preset creation/modification/deletion can be done by site admins (no UI for this yet, so we need to use the django admin panel)

Removed:
* Most of the project stuff, including stuff to do with repos and importing
* The Preset dataclass and all hard-coded presets in the backend
* Babel (.babelrc, @babel/core, babel-loader)

Misc:
* Admins can now delete any scratches from the site UI
* The backend will now return *all* presets as 'available' for a platform
* Organized tests into separate files
* Organized imports for several python source files
* Improved validation for scratch creation, moving some things into the serializer
* Changed wine/wibo to be used even if the host is on Windows, as this provides a smoother and more consistent experience for compilers
* Added PYTHONPATH=backend to .env to help vscode with completion / imports
* Updated .vscode/settings to remove deprecated stuff
* Changed site timezone to "Asia/Tokyo" from "Japan"
* Updated backend deps
* Updated frontend deps

Not in this PR:
* Any sort of UI for creating / editing / deleting presets


backend/coreapp/models/preset.py
```diff
@@ -0,0 +1,21 @@
+from coreapp.models.scratch import LibrariesField
+from django.db import models
+
+
+class Preset(models.Model):
+    name = models.CharField(max_length=100)
+    platform = models.CharField(max_length=100)
+    compiler = models.CharField(max_length=100)
+    assembler_flags = models.TextField(max_length=1000, default="", blank=True)
```
mkst, line 9: only raising for discussion, but why do we distinguish `diff_flags` differently?

backend/coreapp/serializers.py
```diff
@@ -141,6 +188,49 @@ class ScratchCreateSerializer(serializers.Serializer[None]):
...
+            data["platform"] = preset.platform
+            data["compiler"] = preset.compiler
+            data["compiler_flags"] = preset.compiler_flags
+            data["diff_flags"] = preset.diff_flags
+            data["libraries"] = preset.libraries
+        else:
+            if "compiler" not in data:
+                raise serializers.ValidationError("compiler is required")
+            compiler = compilers.from_id(data["compiler"])
+            platform = platforms.from_id(data["platform"])
```
mkst, line 226: We should probably check whether `"platform" not in data`?

We perhaps could/should perform compiler lookup based on compiler_id + platform? Currently I think all compiler_ids are unique, but I'm not sure they have to be (only platform + compiler_id needs to be)?

backend/coreapp/models/preset.py
ethteck: because assembler flags are passed to the assembler directly (for the left side), but diff flags are supposed to control asm-differ functionality itself, I think

backend/coreapp/tests/test_compilation.py
```diff
@@ -0,0 +1,257 @@
...
+from coreapp.flags import Language
+from coreapp.models.scratch import Assembly
+from coreapp.platforms import N64
+from coreapp.tests.common import BaseTestCase, requiresCompiler
+from django.urls import reverse
+from parameterized import parameterized
+from rest_framework import status
+
+
+class CompilationTests(BaseTestCase):
```
mkst, line 23: should we add a 'create-scratch-from-preset' test?

backend/coreapp/serializers.py
ethteck: I have the first part fixed in an upcoming commit

as for the 2nd part, yeah I agree. I'll look at that

backend/coreapp/tests/test_compilation.py
ethteck: it's in test_preset.py

backend/coreapp/serializers.py
ethteck: I think let's address the compiler id + platform thing in a later PR...it's a bit more involved

frontend/src/components/Scratch/ScratchToolbar.tsx
```diff
@@ -173,7 +173,7 @@ function Actions({ isCompiling, compile, scratch, setScratch, setDecompilationTa
                     Fork
                 </button>
             </li>
-            {scratch.owner && userIsYou(scratch.owner) && <li>
+            {((scratch.owner && userIsYou(scratch.owner)) || api.useThisUserIsAdmin()) && <li>
```
nanaian, line 176: Don't call hooks conditionally! We should do
```
const isAdmin = api.useThisUserIsAdmin();
// ...
{((scratch.owner && userIsYou(scratch.owner)) || isAdmin)
```

nanaian (changes requested):

frontend/src/components/Scratch/ScratchToolbar.tsx
ethteck: why?

ethteck: okay, I looked it up and there's information about it here: https://legacy.reactjs.org/docs/hooks-rules.html

nanaian (approved):

PR #900, opened 2023-10-25
Sentry issue fixes

ethteck:
Created this as a result of looking through some of the most common sentry issues that were reported

Things (hopefully) fixed:
* Github login failures weren't being caught and handled
* Scratches with libraries couldn't be forked because of issues with the library field going from the Scratch data into a serializer and then back into a Scratch. I handled this with a hack-ish fix, but I think it's not too bad

There's still more to look into, primarily the stuff with getting user info from GitHub. So we can either handle that here or in a new PR.

### DIFF
--- a/backend/coreapp/models/github.py
+++ b/backend/coreapp/models/github.py
@@ -1,7 +1,6 @@
 import shutil
 import subprocess
 from pathlib import Path
-
 from typing import Any, Optional
 
 import requests
@@ -13,14 +12,15 @@ from django.core.cache import cache
 from django.db import models, transaction
 from django.dispatch import receiver
 from django.utils.timezone import now
-from github import Github, BadCredentialsException
+from github import BadCredentialsException, Github
 from github.NamedUser import NamedUser
 from github.Repository import Repository
+from requests import RequestException
+from requests import ConnectTimeout
 from rest_framework import status
 from rest_framework.exceptions import APIException
 
 from ..middleware import Request
-
 from .profile import Profile
 from .project import Project
 from .scratch import Scratch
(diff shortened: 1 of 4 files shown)


backend/coreapp/models/github.py
```diff
@@ -13,14 +12,15 @@
 from django.db import models, transaction
 from django.dispatch import receiver
 from django.utils.timezone import now
-from github import Github, BadCredentialsException
+from github import BadCredentialsException, Github
 from github.NamedUser import NamedUser
 from github.Repository import Repository
+from requests import RequestException
+from requests import ConnectTimeout
```
mkst, line 19: Combine with above?

backend/coreapp/views/scratch.py
```diff
@@ -461,11 +461,17 @@ def fork(self, request: Request, pk: str) -> Response:
 
         ser = ScratchSerializer(data=fork_data, context={"request": request})
         ser.is_valid(raise_exception=True)
+
+        libraries = [
+            Library(name=lib["name"], version=lib["version"])
```
mkst, line 466: Do we expect different fields or could you pass `**lib`?

mkst (approved):

backend/coreapp/views/scratch.py
ethteck: I'd normally say that I prefer to be explicit, but since we're going from Model -> serializer -> Model, I guess we have bigger problems if those are out of sync ..so sounds good :)

backend/coreapp/views/scratch.py
```diff
@@ -261,9 +261,7 @@ def create_scratch(data: Dict[str, Any], allow_project: bool = False) -> Scratch
     else:
         project_function = None
 
-    libraries = [
```
mkst, line 264: did you change this one, rather than the one I commented? :D

ethteck: oopz...

PR #903, opened 2023-10-30
Fix permuter import

### LINKED ISSUE #904: Permuter import results in 500 error
Works if permuter import.py sends the platform when creating the scratch...

mkst:
Fixes #904. The issue was that the `platform` value was coming back as a `str` in some code paths and a `Platform` in others. Now it should come back as a `Platform`.

Broken test run: https://github.com/decompme/decomp.me/actions/runs/6697436478/job/18197319501

### DIFF
--- a/backend/coreapp/views/scratch.py
+++ b/backend/coreapp/views/scratch.py
@@ -185,11 +185,7 @@ def create_scratch(data: Dict[str, Any], allow_project: bool = False) -> Scratch
     create_ser.is_valid(raise_exception=True)
     data = create_ser.validated_data
 
-    platform: Optional[Platform] = None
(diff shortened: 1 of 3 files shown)


backend/coreapp/views/scratch.py
```diff
@@ -188,7 +188,11 @@ def create_scratch(data: Dict[str, Any], allow_project: bool = False) -> Scratch
     platform: Optional[Platform] = None
     given_platform = data.get("platform")
     if given_platform:
-        platform = platforms.from_id(given_platform)
+        platform = (
```
mkst, line 191: This is absolutely disgusting. Why isn't the type of platform consistent?

PR #908, opened 2023-11-01
parseInt(new_scratch_preset)

### LINKED ISSUE #907: PresetId in new scratch page is not being saved in localStorage correctly.
Screenshot 2023-11-01 114850

Whenever I load up the new scratch page, it says the presetId is "undefined". Which sets the preset to "Custom". Even though I always set it as "Diddy Kong Racing" (preset 27). Not a huge deal, but kinda annoying. This still occurs even if I clear the localStorage data.

I'm using Chrome 118.0.5993.118 on Windows 11.

mkst:
looks good locally (and on Vercel)... the initial preset state is "Custom" but as soon as a platform is selected, the Preset works (for all platforms) - I dont know whether we care enough about that?

Fixes #907.


nanaian:
We should probably handle the initial case where parseInt returns NaN

mkst:
bleurgh.

mkst:
I give up :)

mkst:
nah this change is junk.

ethteck (approved):

mkst:
Right. am finally happy with this. I dont know if there is a better way that would replace the need for the `ready` var?

nanaian (approved):
LGTM for now. A better way to do this in future is probably to make a `useLocalStorageState` hook that wraps `useState` and `useEffect` to load/store into localStorage automatically

PR #932, opened 2023-12-17
Singularise 'libraries' and 'compilers' api endpoints

mkst:
The `/api/compilers` endpoint will still work (i.e. for decomp-permuter looking up decomp.me compilers when import.py is run with --decomp.me).

### DIFF
--- a/backend/coreapp/urls.py
+++ b/backend/coreapp/urls.py
@@ -1,8 +1,8 @@
 from django.urls import path
 
 from coreapp.views import (
-    compilers,
-    libraries,
+    compiler,
+    library,
     platform,
     preset,
     stats,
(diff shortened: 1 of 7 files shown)


nanaian (changes requested):
Can you also update the frontend to use the singular endpoints? Should mostly be a find-replace job

mkst:
> Can you also update the frontend to use the singular endpoints? Should mostly be a find-replace job

lol that was incredibly dumb of me :) done.

Do you know why we dont use SWR for the initial get to compilers+platforms?
```js
export default async function NewScratchPage() {
    const compilers = await get("/compiler")  // y no SWR() ?

    return <main>
        <h1 className="text-2xl font-semibold tracking-tight text-gray-12 md:text-3xl">Start a new scratch</h1>
        <p className="max-w-prose py-3 leading-snug">{DESCRIPTION}</p>
        <NewScratchForm serverCompilers={compilers} />
    </main>
}
```

nanaian:
That's a [server component](https://react.dev/reference/react/use-server) so we can just `await` a request like that. SWR is for client components only

Issue #215, opened 2021-11-22
Upload object file as base


ethteck:
allow the user to upload an object file from which to get the target asm

PR #943, opened 2024-01-15
Support binary `target_obj` in scratch creation

encounter:
Also support `--disassemble=` in `diff_flags`, so the scratch can be limited to a single function within the object.

Example:
[object.zip](https://github.com/decompme/decomp.me/files/13943393/object.zip)
```
curl -v http://127.0.0.1:8000/api/scratch \
  -F compiler=mwcc_242_81 \
  -F platform=gc_wii \
  -F compiler_flags=-lang=c++ \
  -F diff_label=__copy_longs_unaligned \
  -F target_obj=@mem_funcs.o \
  -F 'diff_flags=["--disassemble=__copy_longs_unaligned"]'
  -F context= \
  -F source_code= \
```

Note that the object contains 3 functions, but the final scratch is properly filtered down to just `__copy_longs_unaligned`, because of the value passed for `diff_flags`.

Resolves https://github.com/decompme/decomp.me/issues/215


backend/coreapp/tests/common.py
```diff
@@ -18,7 +18,7 @@ def requiresCompiler(*compilers: Compiler) -> Callable[..., Any]:
 
 class BaseTestCase(APITestCase):
     # Create a scratch and return it as a DB object
-    def create_scratch(self, partial: Dict[str, str]) -> Scratch:
+    def create_scratch(self, partial: Dict[str, Any]) -> Scratch:
```
nanaian, line 21: Why was this needed?

PR #945, opened 2024-01-17
Use random token for claiming scratches

### LINKED ISSUE #553: Claim button gone for scratches created anonymously
You can currently ctrl+s to save, manually issue a /claim call, or fork to work around this

### LINKED ISSUE #559: Have url returned making an unclaimed scratch contain a claim key in query param
Scratches should not be claimable without a unique token
This should resolve #553

encounter:
On scratch `create` or `fork`, a new `claim_token` is returned, and is now required to successfully claim a scratch.

This is used in both:
- `POST /api/scratch/:slug/claim?token=` Existing endpoint with new token parameter.
- Navigate to `/scratch/:slug/claim?token=` in the browser. This will claim the scratch and redirect to the scratch page.

The new browser route is useful for tools that create scratches via the API and then want to open the scratch in the user's browser and have it automatically claimed.

Fixes #553 
Fixes #559


frontend/src/app/scratch/[slug]/claim/page.tsx
```diff
@@ -0,0 +1,32 @@
...
+}) {
+    const router = useRouter()
+
+    // The POST request must happen on the client so
+    // that the Django session cookie is present.
+    const effectRan = useRef(false)
+    useEffect(() => {
+        if (!effectRan.current) {
+            post(`/scratch/${params.slug}/claim`, { token: searchParams.token })
+                .catch(err => console.error(err))
```
Comment (nanaian) [inline frontend/src/app/scratch/[slug]/claim/page.tsx:22]: Need to do proper error handling; if there's an error (e.g. the scratch is already claimed) then this page will appear to load forever

frontend/src/lib/api/types.ts
```diff
@@ -52,6 +52,7 @@ export interface Scratch extends TerseScratch {
     source_code: string
     context: string
     diff_label: string
+    claim_token: string | null
```
nanaian, line 55: This should be able to be undefined since ScratchSerializer doesn't include it

frontend/src/lib/api.ts
```diff
@@ -101,7 +101,9 @@ export function useSaveScratch(localScratch: Scratch): () => Promise<Scratch> {
 }
 
 export async function claimScratch(scratch: Scratch): Promise<void> {
-    const { success } = await post(`${scratchUrl(scratch)}/claim`, {})
+    const { success } = await post(`${scratchUrl(scratch)}/claim`, {
+        token: scratch.claim_token,
```
nanaian, line 105: Correct me if I'm wrong but it seems this will be `undefined` when scratch didn't come from create_scratch, e.g. if you just visit an unclaimed scratch

nanaian (changes requested):
I like this approach. I think we should remove claiming from the fuzzy save logic and update tools that work with decomp.me such as decomp-permuter to present a link with a new claim endpoint using the token they got from the API.

Perhaps claim_token would be better off as a separate table (scratch_id, claim_token) where a row only exists if the scratch is claimable. This avoids having to do the awkward `pop` when forking, and avoids the migration adding a claim token to all existing scratches regardless of whether they need one.

frontend/src/lib/api.ts
nanaian: If https://github.com/decompme/decomp.me/pull/945#discussion_r1454839593 is applied, then throw an error if the claim token is missing on `scratch`

frontend/src/lib/api/types.ts
nanaian: Alternatively do `interface ClaimableScratch extends Scratch` like the backend

frontend/src/app/scratch/[slug]/claim/page.tsx
Comment (encounter) [reply frontend/src/app/scratch/[slug]/claim/page.tsx:22]: That's what the `.finally` block handles!

ethteck:
I might be missing something, but I believe this change warrants also updating https://github.com/simonlindholm/decomp-permuter to give the user the claim link

frontend/src/lib/api.ts
encounter: Updated TS types for this

frontend/src/lib/api/types.ts
encounter: Done

encounter:
Updated, still need to fix tests...

> I think we should remove claiming from the fuzzy save logic

Done

> This avoids having to do the awkward `pop` when forking

This was resolved by fixing serialization logic. Before, a `"claim_token": None` would sneak into the request data, making it not generate one.

> and avoids the migration adding a claim token to all existing scratches regardless of whether they need one.

This was discussed in Discord, but the migration should leave all existing rows with a null claim_token.

> update tools that work with decomp.me such as decomp-permuter to present a link with a new claim endpoint using the token they got from the API.

> I might be missing something, but I believe this change warrants also updating https://github.com/simonlindholm/decomp-permuter to give the user the claim link

Yes to both 🙂

frontend/src/app/scratch/[slug]/claim/page.tsx
Comment (nanaian) [reply frontend/src/app/scratch/[slug]/claim/page.tsx:22]: If the claim fails I think we should display the error rather than silently redirect

frontend/src/app/scratch/[slug]/claim/page.tsx
```diff
@@ -0,0 +1,32 @@
...
+                .finally(() => router.push(`/scratch/${params.slug}`))
```
Comment (nanaian) [inline frontend/src/app/scratch/[slug]/claim/page.tsx:23]: Perhaps this should be `replace`

nanaian:
An alternative might be to add a query param to the scratch page like `?claim=token`. Then we don't have to fuss about with redirects and we could show a banner on success/fail?

encounter:
That was my initial approach, but it required fully loading the page (SSR), before the client could then go hit the claim endpoint. Then it'd have to reload the scratch info and the UI would go from unclaimed to claimed. Didn't seem optimal, I think it's better to claim before loading any scratch info whatsoever.

encounter:
Everything should be fixed up now.

PR for decomp-permuter: https://github.com/simonlindholm/decomp-permuter/pull/165

PR #953, opened 2024-02-07
Add current.o to export zip

simonlindholm:
Does this seem reasonable? It makes exports slightly slower, but I imagine the feature isn't used much (?). I've needed this a couple of times for debugging.


ethteck:
yeh seems fine to me!

PR #956, opened 2024-02-09
Stop relying on GitHub's API so much

ethteck:
Hopefully will fix perf issues we've been having lately and also reduces network use / complexity of the site


mkst (approved):

backend/coreapp/serializers.py
```diff
@@ -48,18 +47,14 @@ def serialize_profile(
...
         if small:
             return small_obj
 
         return {
             **small_obj,
-            "email": user.email,
-            "name": github_details.name if github_details else user.username,
-            "github_api_url": github_details.url if github_details else None,
-            "github_html_url": github_details.html_url if github_details else None,
+            "github_id": gh_user.github_id if gh_user else None,
```
bates64, line 57: It probs makes sense to get rid of the small/large user distinction in future if there's only one extra field

frontend/src/lib/api/urls.ts
```diff
@@ -8,6 +8,15 @@ export function userUrl(user: User): string {
     return `/users/${user.username}`
 }
 
+export function userAvatarUrl(user: User): string | null {
+    return user.github_id && `https://avatars.githubusercontent.com/u/${user.github_id}`
+}
+
+// Only returns a URL if the user has a GitHub ID
+export function userGithubHtmlUrl(user: User): string | null {
+    return user.github_id && `https://github.com/${user.username}`
```
bates64, line 17: Better to use `??` over `&&` I think

ethteck: that makes it return the id if it's present apparently, which isn't the behavior I want here

PR #1102, opened 2024-02-25
Add a link to presets on the scratch list

polybiusproxy:


frontend/src/components/ScratchList.tsx
```diff
@@ -118,19 +126,25 @@ export function ScratchItemNoOwner({ scratch }: { scratch: api.TerseScratch }) {
     const compilerName = compilersTranslation.t(scratch.compiler)
     const serverPresets = api.usePlatforms()[scratch.platform].presets
     const presetName = serverPresets.find(p => p.id === scratch.preset)?.name
+    const preset = serverPresets.find(p => p.id === scratch.preset)
```
ethteck, line 129: I believe if you put this preset line above the presetName one, presetName can just be `preset?.name` rather than doing this find() twice

frontend/src/components/ScratchList.tsx
```diff
@@ -79,18 +81,24 @@ export function getMatchPercentString(scratch: api.TerseScratch) {
     return matchPercentString
 }
 
-export function ScratchItem({ scratch, children }: { scratch: api.TerseScratch, children?: ReactNode }) {
+export function ScratchItem({ scratch, children, noLink } : { scratch: api.TerseScratch, children?: ReactNode, noLink?: boolean }) {
     const compilersTranslation = useTranslation("compilers")
     const compilerName = compilersTranslation.t(scratch.compiler as any)
     const serverPresets = api.usePlatforms()[scratch.platform].presets
     const presetName = serverPresets.find(p => p.id === scratch.preset)?.name
+    const preset = serverPresets.find(p => p.id === scratch.preset)
```
ethteck, line 89: here2

frontend/src/components/ScratchList.tsx
```diff
@@ -79,18 +81,24 @@ export function getMatchPercentString(scratch: api.TerseScratch) {
     return matchPercentString
 }
 
-export function ScratchItem({ scratch, children }: { scratch: api.TerseScratch, children?: ReactNode }) {
+export function ScratchItem({ scratch, children, noLink } : { scratch: api.TerseScratch, children?: ReactNode, noLink?: boolean }) {
```
ethteck, line 84: I think it makes more sense to call this property isLink (and invert the logic)

.vscode/settings.json
```diff
@@ -40,4 +40,4 @@
         "--config-file=backend/mypy.ini"
     ],
     "editor.formatOnSave": true
-}
+}
```
ethteck, line 43: I think an empty line should be added here. or maybe it wasn't here and you added it? github diffs for this always confuse me

ethteck (changes requested):
thanks for this! a couple change requests

.vscode/settings.json
polybiusproxy: Seems like it was added by the formatter.

PR #1106, opened 2024-02-27
Make MWCCCompiler more granular for each platform

### LINKED ISSUE #1105: Make MWCCCompiler class more granular

The MWCCCompiler class is used by PSP, PS2 and GC/Wii platforms, but the compilers have different options - so it makes sense to split MWCCCompiler into platform specific versions.

We shouldn't show compiler options that arent applicable to the compiler.

n/a

**Additional context**
n/a

mkst:
This is a best effort based on running one of each compiler with the `--help` argument. We can tweak it as necessary.

Closes #1105.

### DIFF
--- a/backend/coreapp/flags.py
+++ b/backend/coreapp/flags.py
@@ -204,7 +204,6 @@ COMMON_MWCC_NDS_ARM9_FLAGS = COMMON_MWCC_FLAGS + [
     Checkbox(id="mwcc_rostr", flag="-rostr"),
     Checkbox(id="mwcc_enc_sjis", flag="-enc SJIS"),
 ]
-print(COMMON_MWCC_NDS_ARM9_FLAGS)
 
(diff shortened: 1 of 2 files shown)


backend/coreapp/flags.py
```diff
@@ -189,15 +188,51 @@ def to_json(self) -> Dict[str, Union[str, List[str]]]:
...
+        flags=[
+            "-fp soft",
+            "-fp vfpv1",
+            "-fp vfpv2",
+        ],
+    ),
+    Checkbox(id="mwcc_rostr", flag="-rostr"),
+    Checkbox(id="mwcc_enc_sjis", flag="-enc SJIS"),
+]
+print(COMMON_MWCC_NDS_ARM9_FLAGS)
```
ethteck, line 207: stray print?

PR #1107, opened 2024-03-02
Use custom powerpc-eabi binutils for GC/Wii

encounter:
Source: https://github.com/encounter/gc-wii-binutils

Right now there's only one patch to fix mftb instructions, but it's an issue that pops up pretty frequently.


encounter:
Hmm, nsjail doesn't seem to like the binaries. Is it because they're statically linked?

ethteck:
interesting.. it's only for the docker build, so it must be some permission thing in there I assume

mkst:
Thankfully an easy fix :+1:

mkst (approved):

Issue #772, opened 2023-05-21
Shortcut viewer / editor


ethteck:
Particularly after #639, it seems like a no-brainer to have a way to view / modify shortcuts on the site. This can be added as a new Settings page.

jesusb25:
Hi! I'm a student in the CodePath AI301 program and I'd like to work on this issue. I'll start by exploring the codebase this week. Any pointers on where to start in the codebase would be appreciated!

PR #1134, opened 2024-03-12
Add setting for vim bindings to be used

### LINKED ISSUE #1130: vim keybindings to the editor maybe?
I've gotten too used to vim keybindings 

A setting to enable vim keybindings (would be cool I think).

I've tried learning the Official Decomp.Me Shortcuts™, but muscle memory is a beast :disappointed: 

**Additional context**
idk if it's possible but it would be cool tho :pray:

wearrrrr:
Resolves #1130 by adding vim bindings as an option inside of `/settings/editor`, apologies in advance if I didn't make the most optimal react code, I have worked with it very little. Doesn't seem to cause any lag for me, but I believe that it might be rerendering too much

### DIFF
--- a/frontend/src/components/Scratch/Scratch.tsx
+++ b/frontend/src/components/Scratch/Scratch.tsx
@@ -2,12 +2,13 @@ import { useEffect, useReducer, useRef, useState } from "react"
 
 import { cpp } from "@codemirror/lang-cpp"
 import { EditorView } from "@codemirror/view"
+import { vim } from "@replit/codemirror-vim"
 
 import * as api from "@/lib/api"
 import basicSetup from "@/lib/codemirror/basic-setup"
 import useCompareExtension from "@/lib/codemirror/useCompareExtension"
 import { useSize } from "@/lib/hooks"
-import { useAutoRecompileSetting, useAutoRecompileDelaySetting, useLanguageServerEnabled } from "@/lib/settings"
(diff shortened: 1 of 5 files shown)


ethteck (approved):
I can't speak to the UX of using this on the site, but the code looks good! Have you tested this? My only concern is that vim keybindings would conflict with others we have on the site. We probably should prioritize #772

wearrrrr:
I did have the original creator of the issue test the changes, which they reported to be good, except for Ctrl + F conflicting with Decomp.me's keybind for find, I definitely agree that a custom hotkey system would be great and probably not *super* difficult

frontend/src/components/Scratch/Scratch.tsx
```diff
@@ -169,6 +169,15 @@ export default function Scratch({
         incrementValueVersion()
     }, [scratch.slug, scratch.last_updated])
 
+    const cmExtensions = [...CODEMIRROR_EXTENSIONS, sourceCompareExtension]
+    const cmExtensionsRef = useRef(cmExtensions)
+    useEffect(() => {
+        if (useVim == true) {
+            cmExtensionsRef.current = [...cmExtensionsRef.current, vim()]
+        }
+    })
```
simonlindholm, line 178: This setup with useEffect and refs confuses me, why is it not just `let cmExtensions = [...CODE_MIRROR_EXTENSIONS, sourceCompareExtension; if (useVim) cmExtensions.push(vim())`?

In particular it looks like this useEffect callback will fire any time this component re-renders, and continuously append `vim()` entries to the array, never removing them. Also, this will happen with the wrong timing: after render, rather than immediately.

PR #1138, opened 2024-03-16
Add 'has_decompiler' to Platform

### LINKED ISSUE #492: Expose supported decompilers in platforms under /compilers API
We can provide a list of platforms for which we have a decompiler and then disable the decompile button for scratches of platforms for which we lack a decompiler

mkst:
Closes #492.

I think I've got them all?

- N64
- IRIX
- PS1
- PSP
- PS2
- GC_WII


wearrrrr:
Seems correct to me!

PR #1139, opened 2024-03-16
Swap usePlatforms for usePresets/usePreset

mkst:
The *upside* to this change is that rather than getting all compilers + platforms + presets etc via the `/compiler`, we grab each preset as required.

```sh
$ curl --silent https://decomp.me/api/compiler?format=json | wc -c
173635
```

The *downside* to this change is that rather than getting all compilers + platforms + presets etc via the `/compiler`, we grab each preset as required. 

```sh
$ curl --silent "https://decomp.me/api/preset/21?format=json" | wc -c
244
```

I still think this is much better in terms of amount of data being sent & processed - but it is more requests to the backend (I know there is a maximum of 6 outstanding requests per tab, does this also apply to `fetch`)?

~~Gonna see how the vercel deployment looks...~~

**Edit:**

Performance seems good to me... When loading the front page we load ~20 scratches, worst-case scenario would be 20 *different* presets, but currently there are only 5 requests needed (due to scratches without any preset, as well as a number of scratches using the same preset). Loading the [platform](https://frontend-66fcwve8y-decompme.vercel.app/platform/n64) page for n64 only requires 3 calls to the presets endpoint, so I think it's OK.

**Bugs:**

- ```Scratch.preset == '97' but no preset with that id was found.``` :green_circle: Fixed!

### DIFF
--- a/frontend/src/components/ScratchList.tsx
+++ b/frontend/src/components/ScratchList.tsx
@@ -82,9 +82,8 @@ export function getMatchPercentString(scratch: api.TerseScratch) {
 export function ScratchItem({ scratch, children } : { scratch: api.TerseScratch, children?: ReactNode }) {
     const compilersTranslation = useTranslation("compilers")
     const compilerName = compilersTranslation.t(scratch.compiler as any)
-    const serverPresets = api.usePlatforms()[scratch.platform].presets
     const matchPercentString = getMatchPercentString(scratch)
-    const preset = serverPresets.find(p => p.id === scratch.preset)
+    const preset = api.usePreset(scratch.preset)
     const presetName = preset?.name
 
     const presetOrCompiler = presetName ?
@@ -124,9 +123,8 @@ export function ScratchItem({ scratch, children } : { scratch: api.TerseScratch,
 export function ScratchItemNoOwner({ scratch }: { scratch: api.TerseScratch }) {
     const compilersTranslation = useTranslation("compilers")
     const compilerName = compilersTranslation.t(scratch.compiler)
-    const serverPresets = api.usePlatforms()[scratch.platform].presets
     const matchPercentString = getMatchPercentString(scratch)
-    const preset = serverPresets.find(p => p.id === scratch.preset)
+    const preset = api.usePreset(scratch.preset)
     const presetName = preset?.name
 
     const presetOrCompiler = presetName ?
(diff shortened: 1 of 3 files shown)


ethteck (approved):

frontend/src/lib/api.ts
```diff
@@ -288,6 +277,24 @@ export function useLibraries(): LibraryVersions[] {
     return data.libraries
 }
 
+export function usePresets(platform: string | undefined): Preset[] {
+    const getByPlatform = ([url, platform]) => {
+        return get(url && platform && `${url}?platform=${platform}&page_size=100`)
```
ethteck, line 282: I just realized the site will cease to show any more than 100 presets with this change. I think what we eventually want is a searchable thing with auto-populated results, and then an "Add new..." for creating a new preset 

For now, instead of usePresets, we might want to keep what we had before, so we're sure not to omit any presets from the dropdown

mkst: It's 100 presets *per platform*, which I think will be fine until we get around to revamping it (i.e. when we let users create presets)

bates64:
With HTTP/2 doing lots of requests at once shouldn't be too bad. Particularly if we start caching the frontend / doing async django in future.

frontend/src/components/ScratchList.tsx
```diff
@@ -82,9 +82,8 @@ export function getMatchPercentString(scratch: api.TerseScratch) {
 export function ScratchItem({ scratch, children } : { scratch: api.TerseScratch, children?: ReactNode }) {
     const compilersTranslation = useTranslation("compilers")
     const compilerName = compilersTranslation.t(scratch.compiler as any)
-    const serverPresets = api.usePlatforms()[scratch.platform].presets
     const matchPercentString = getMatchPercentString(scratch)
-    const preset = serverPresets.find(p => p.id === scratch.preset)
+    const preset = api.usePreset(scratch.preset)
     const presetName = preset?.name
 
     const presetOrCompiler = presetName ?
```
bates64, line 89: Consider changing L92 to use a loading skeleton element in place of the preset name - it looks a bit iffy to flash from compiler name to preset name once usePreset finishes loading

bates64: Actually, there will be a flash regardless because of the width and there being text to the right. Best thing to do is to view transition this I think, regardless of whether its a skeleton or text.

Shouldn't need to block this PR though, we should make an issue out of it if it's nontrivial

frontend/src/lib/api.ts
mkst: Assert that the length of presets is < 100? Currently we have 43 presets for N64 (which I'm assuming is the most popular platform). I'll create an issue to make the Presets a searchable drop-down - I feel it's likely we'll resolve that before we hit 100 presets for any individual platform.

frontend/src/components/ScratchList.tsx
mkst: Do you mind creating an issue so it's not forgotten? I don't know what "view transition this" means :(

frontend/src/lib/api.ts
mkst: Created https://github.com/decompme/decomp.me/issues/1152.

PR #1142, opened 2024-03-16
Tweak vim bindings PR

mkst:


wearrrrr:
LGTM, the vim bindings should also be set here, something I forgot to do in the original PR. https://github.com/decompme/decomp.me/blob/aec3001b8d71ed85a0a93a3b0ddd69f01709247d/frontend/src/components/Scratch/Scratch.tsx#L243

mkst:
Done. I didn't realise what an improvement it would be to have these vim bindings - so good!

frontend/src/components/Scratch/Scratch.tsx
```diff
@@ -182,13 +182,9 @@ export default function Scratch({
     }, [scratch.slug, scratch.last_updated])
 
     const cmExtensions = [...CODEMIRROR_EXTENSIONS, sourceCompareExtension]
+    if (useVim)
+        cmExtensions.push(vim())
     const cmExtensionsRef = useRef(cmExtensions)
```
simonlindholm, line 187: Why this ref? Feels like it should make the list of extensions never change from the first render, since the argument to useRef just serves as a default value

mkst: So just use `cmExtensions` instead of creating `cmExtensionsRef` ?

simonlindholm: Yeah

PR #1146, opened 2024-03-17
Fixup vim refs (also fix context comparison)

mkst:
Somewhere along the line we stopped passing `contextCompareExtension` to the context pane.


simonlindholm:
The vim part looks good to me assuming it works!

PR #1147, opened 2024-03-17
Three-way diffs

### LINKED ISSUE #93: Three-way diffs
Easiest done JS-side. I guess we'll want to diff against the asm of the code that's been saved?

simonlindholm:
Fixes #93, by doing simple client-side diffing. Same algorithm and UX as asm-differ, with the latest saved version of the scratch being used as the three-way base. Three-way diffing is turned on by clicking the small button saying "2" or "3" to the RHS of the diff header -- I think it makes sense to put it there instead of in scratch or profile settings, since it's somewhat of a per-session preference.

I would very much appreciate feedback on this! Since it's a pure client-side feature the vercel PR deployment should work for testing.


frontend/src/components/Diff/Diff.tsx
```diff
@@ -185,59 +185,86 @@ function DiffBody({ diff, fontSize }: { diff: api.DiffOutput, fontSize: number |
...
 
+function ThreeWayToggleButton({ enabled, setEnabled }: { enabled: boolean, setEnabled: (enabled: boolean) => void }) {
+    return <button
+        className={styles.threeWayToggle}
+        onClick={() => {
+            setEnabled(!enabled)
+        }}
+        title={enabled ? "Disable three-way diffing" : "Enable three-way diffing"}
+    >
+        {enabled ? "3" : "2"}
```
bates64, line 196: Possibly needs some iteration, like an icon or something. On mobile you can't ever see the title

frontend/src/lib/interdiff.ts
```diff
@@ -0,0 +1,101 @@
...
+        }
+        while (ci != cs.length)
+            addMatching(cs[ci++], ps[pi++])
+    }
+
+    const currChunks = chunkDiffLines(curr.rows)
+    const prevChunks = chunkDiffLines(prev.rows)
+    if (currChunks.chunks.length !== prevChunks.chunks.length) {
+        // This should logically never happen, since the two diffs are based on
+        // the same target.
```
bates64, line 82: Warn?

bates64 (commented):
Awesome work, and enormously valuable! Thanks Simon. Couple comments but on the whole LGTM. Vercel deployment works on mobile too.

frontend/src/components/Diff/Diff.tsx
simonlindholm: I would love some help with this, my design skills are terrible.

bates64: Maybe we use this icon? https://primer.style/foundations/icons/git-merge-16

frontend/src/lib/interdiff.ts
simonlindholm: Good point. I can add a console.warn; were you thinking some visual indication, and if so do you have tips on how to do it?

bates64: Just a console warn is fine :shipit:

simonlindholm: I added a console.warn with the two diff objs being logged to the console; I think that's enough to troubleshoot problems if they come up (I don't expect them to, haven't had reports on asm-differ). This condition will become user-triggerable if we ever make target asm editable, btw, so I don't want to make it too hard an error.

mkst:
Love it. Nice one Simon!
image

PR #1150, opened 2024-03-20
Toggle problems by clicking anywhere on the bar

Daniel-McCarthy:
## Description

Previously the "Problems" bar could only be toggled by clicking the small icon on the very far right. It would be a more pleasant of a user experience to be able to simply click the bar anywhere to toggle it, as it is much quicker and requires less precision.

This stretches the button to fit the full bar while keeping the spacing of the text an arrow icon as before.

## Demonstration
Peek_2024-03-19_18-37


bates64 (approved):
"No problems"! Great change, thanks for the PR!

PR #1164, opened 2024-03-24
Fix indentation when braces are on new lines

WhenGryphonsFly:
Currently, there is an inconsistency with the indentation of braces. When used in an if-statement, the braces are in-line with the if statement. In almost every other context (for loops, while loops, do-while loops, switch statements, and try-catch statements), the braces are indented 1 level, and then the contents of the braces are further indented.
image

This PR makes this indentation consistent, standardizing on braces always being in-line with their parent statements.
image

This is accomplished by incorporating [`codemirror/lang-cpp/src/cpp.ts`](https://github.com/codemirror/lang-cpp/blob/main/src/cpp.ts) into this repository and making the necessary changes. This option was chosen over sending a PR to upstream because this is an issue of style, rather than one of function.

I am open to further changes, such as moving the file elsewhere or modifying the style standardized on.

Note: based on my testing, there are no observable changes for those who put braces on the same line as the parent statement. This is one of the reasons why the style of switch statements was left unchanged. (The other reason is that it is surprisingly difficult to do so.)


WhenGryphonsFly:
Also, if anyone knows how to run the frontend linter in docker I would appreciate it

mkst:
It should be something like:
```docker exec decompme-frontend-1 yarn lint```
You can get the name of the front end container with `docker ps`

ethteck (approved):

PR #1165, opened 2024-03-25
Change `mwcc_enum_int` checkbox to a `FlagSet`

AngheloAlf:
`-enum int` is the default on mwcps2, so the current checkbox doesn't make much sense.

I don't know how this would affect existing presets.


frontend/src/lib/i18n/locales/en/compilers.json
```diff
@@ -112,7 +112,6 @@
     "prodg_35": "ProDG 3.5 (gcc 2.95.2)",
     "prodg_37": "ProDG 3.7 (gcc 2.95.2)",
     "prodg_393": "ProDG 3.9.3 (gcc 2.95.3)",
-
```
mkst, line 115: I'm in two minds about all the whitespace getting dropped:
1. lose the whitespace: it happens to me when i save the file in vscode, so i end up making changes using `vim`
2. keep the whitespace: it's nicer to have a separator between the sections

Can someone else weigh in?

AngheloAlf: I personally think those spaces are nice. Maybe remove the auto format on save for this file?

ethteck: rofl you are the one who suggested the change! I'm silly

AngheloAlf: I couldn't find how to disable auto format on a single file, but I found how to save without formatting the file, do `ctrl-k` + `s`

PR #1167, opened 2024-03-25
More ABI register name options for MIPS

AngheloAlf:
Changes the old `ABI FPR names` checkbox into a dropdown that contains all the ABIs supported by objdump.

Initially this was implemented as a checkbox because at the time we only cared about platforms that used the O32 ABI (PS1 and N64), but now that platforms that use other ABIs are getting more love (like PS2 which uses EABI) I think we should properly support them.

My understanding is that resets that had the old checkbox enabled should not be affected by this change since the underlying flag is still the same (`-Mreg-names=32`).

I would recommend all PS2 projects to use the "ABI N32 and EABI" option from now on, it should give more reasonable register names, specially for functions with lots of args

image


ethteck (changes requested):
LGTM! But as someone said (I think) in discord, would you mind undoing the removal of empty lines on the translation file? We really need to disable auto-formatting for it..

ethteck:
lol it was you https://github.com/decompme/decomp.me/pull/1165#discussion_r1538199744 my b

AngheloAlf:
rofl

PR #1169, opened 2024-03-25
Add 'spellCheck={false}' to all relevant <input> components

### LINKED ISSUE #1168: [BUG] Inputs show spell-checking markers
Inputs in the option panel contain red underlines. We should go through all `inputs` and add `spellcheck="false"` to any non-prose input fields.

image

mkst:
Closes #1168. I dont have any spellchecking on the live site (using Chrome), so can someone confirm this change does what it should?

### DIFF
--- a/frontend/src/app/(navfooter)/new/NewScratchForm.tsx
+++ b/frontend/src/app/(navfooter)/new/NewScratchForm.tsx
@@ -281,7 +281,7 @@ export default function NewScratchForm({ serverCompilers }: {
                 className={styles.textInput}
                 autoCorrect="off"
                 autoCapitalize="off"
-                spellCheck="false"
+                spellCheck={false}
             />
         </div>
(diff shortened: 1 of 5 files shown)


marijnvdwerf:
**Checked**
- ✅ Scratch > Compiler options
- ✅ New Scratch > Diff label (already worked)
- ✅ Settings > Font Family

**Not working**
- ❌ Scratch > Diff options > Diff label

I have no idea where I should find the `SearchBox` component

mkst:
Search is top left on the main page.

Re: diff label, the code is there, this is what I see when i inspect the diff label input, `spellcheck="false"` is set:
image

marijnvdwerf:
Ah, that one seems to already work properly on prod. I guess `<input type="search">` acts differently.

`SearchItem` doesn't appear to be used at all in the codebase?

PR #1175, opened 2024-03-27
Have NetworkErrorPage use `window.location.reload` instead of `reset`

### LINKED ISSUE #870: "Try again" button on "We're having some trouble reaching the backend" page seems ineffective
We now have a nicer page that pops up when the backend goes down, but unfortunately the "Try again" button seems to do nothing, even when the site is back up.

ConorBobbleHat:
Fixes #870.

I've left `UnexpectedErrorPage` as using `reset` - it feels like a pure client-side rerender might be more appropriate as a recovery option there? I'm open to arguments the other way, too.

### DIFF
--- a/frontend/src/app/error.tsx
+++ b/frontend/src/app/error.tsx
@@ -2,8 +2,6 @@
 
 import { useEffect } from "react"
 
-import { useRouter } from "next/navigation"
-
 import { SyncIcon } from "@primer/octicons-react"
 
 import Button from "@/components/Button"
--- a/frontend/src/app/error.tsx
+++ b/frontend/src/app/error.tsx
@@ -2,6 +2,8 @@
 
 import { useEffect } from "react"
 
+import { useRouter } from "next/navigation"
+
 import { SyncIcon } from "@primer/octicons-react"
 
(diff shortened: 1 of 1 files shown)


frontend/src/app/error.tsx
```diff
@@ -11,7 +11,7 @@ import { RequestFailedError } from "@/lib/api"
 
 type ErrorPageProps = {error: Error, reset: () => void };
 
-function NetworkErrorPage({ error, reset }: ErrorPageProps) {
+function NetworkErrorPage({ error, reset: _reset }: ErrorPageProps) {
```
marijnvdwerf, line 14: Is there any reason for destructuring the `reset` field of `ErrorPageProps` when it's not used?

ethteck: it's used in `UnexpectedErrorPage` further down, and both take the same props

marijnvdwerf: But we only need to destructure the props that we use, right? Or is there any magic happening?

ethteck: ah not sure, guess I misunderstood

Issue #1113, opened 2024-03-09
Add opengraph metadata for decomp.me homepage


mkst:
Currently links to `decomp.me` itself are pretty boring, e.g.

![image](https://github.com/decompme/decomp.me/assets/22226349/157d43a5-75ff-4855-9fdb-7d90026b4bab)

**Describe the solution you'd like**
Off the back of the improved opengraph image for individual scratches, we should take the opportunity to provide a _good_ embed for anyone creating a link to decomp.me itself.

> **decomp.me**
> A collaborative reverse-engineering platform for working on decompilation projects with others to learn about how your favorite games work.
> 123,456 unique scratches created to date.

We could use opengraph-image here too if we came up with a design :)

mkst:
This is done.

PR #1203, opened 2024-04-18
Added title, description, and og-image to og:metadata (#1113)

conorgolden1:
Below is what the og link looks like on the facebook debugger: (I don't know how to get the facebook debugger to get the image even with a tunnel)

Screenshot from 2024-04-18 16-47-17

Here is an example of what is currently generated from the platformIcon list on the og-image:

image

Here is an example as to how the og-image will readjust automatically to the inclusion of more icons in the list:

image


conorgolden1:
Updated design:

image

conorgolden1:
Updated design:

image

conorgolden1:
Final:

image

PR #1227, opened 2024-05-08
Restrained max width on new scratch

### LINKED ISSUE #663: New scratch code areas have unconstrained width
https://user-images.githubusercontent.com/9429556/214747516-8dfad56b-0a68-4d80-a9d0-fffa5d0eb1a7.mp4

conorgolden1:
Fixes #663


conorgolden1:
The width here is defaulted by the automatic margins on the left and right in the parent div. Since the automatic margins are restricting the width the center content doesn't inherently reach the max width. I just adjusted the max width the about default rendering that is expected. Although I do think we should allow for better scaling based on screen size that is a different issue altogether and one that should be fixed with a mobile view implementation.

Issue #648, opened 2023-01-05
Allow users to create/update presets on site


bates64:
Currently, the process for creating presets requires forking this repo, editing backend code, and creating a pull request. If we put presets in the database and allowed users to edit them, that would make it easier and not require redeploys.

ethteck:
Partially done in #897

axel7083:
I took a deep dive into how presets are used, and here are the steps required to allow user to create/update presets

## Adding `owner` foreign key to `preset` table

In a similar maner than scratch are storing the ownership, the `preset` table should add an `owner` column.

https://github.com/decompme/decomp.me/blob/ed910d70a609523a4087b4f13e61cf0213843ad7/backend/coreapp/models/scratch.py#L117

> ℹ️ We would keep `null` for the _system_ presets, being the one provided by the site, and not owned.

> ⚠️ I would not replicate the _claimable_ mechanism of the scratches, where anonymous user could create presets, I would only allow authentificated user to be able to create presets, so we avoid the claimable stuff.

## Updating permissions of the `PresetViewSet`

Currently the permissions are very limited, and only admins can create new presets.

https://github.com/decompme/decomp.me/blob/f61668a77b17ef5d4ffc15d14d1d15c7f66c8a5d/backend/coreapp/views/preset.py#L29

We would have to change this, to allow authentificated user to create (POST) and manage (PUT/DELETE) their presets. 

> ℹ️ This is not a difficult change.

## Manage how presets are dispayed

Currently presets are included in the following endpoints

- `/api/compiler`
- `/api/preset`

But how do you filter the presets, does all user have access to all presets of everyone ? This has a design flaw, as let's say someone is creating 500 presets, user would have some issue finding the one they would probably expect.

We could for example, filter the presets, and only display the one by the system (without ownership), and the one of the user. But this would makes hard to share the presets. We could also imagine, that user should duplicates/fork presets of user, if they want to use them.

This is the point where I am not sure of the design to adopt.

## Frontend

![image](https://github.com/decompme/decomp.me/assets/42176370/7f6f089a-4b92-4539-b586-c14592511d23)

The dropdown should probably be migrated to some searchable component, since a lot of presets could be shown here. 

## Additional 

I would after the table to add a UNIQUE contraint on the name to avoid confusion.

https://github.com/decompme/decomp.me/blob/c80abfaa0502018d15444aee6481cbb1978e80fe/backend/coreapp/models/preset.py#L10

ethteck:
[...]

PR #1228, opened 2024-05-09
feat(backend): allowing auth users to create presets

axel7083:
## Description

This pull request allows users to create/update/delete presets. 

> ℹ️ This PR is only adding the backend logic to ease review.

## Related issue

Partially fixes https://github.com/decompme/decomp.me/issues/648

## Tests

- [x] Unit tests has been added to ensure expected behaviour

### DIFF
--- a/backend/coreapp/migrations/0042_migrate_presets_to_db.py
+++ b/backend/coreapp/migrations/0042_migrate_presets_to_db.py
@@ -823,6 +823,7 @@ def create_presets(apps: Apps, schema_editor: BaseDatabaseSchemaEditor) -> None:
 class Migration(migrations.Migration):
     dependencies = [
         ("coreapp", "0041_remove_project_and_github_stuff"),
+        ("coreapp", "0054_presets_owner"),
     ]
 
     operations = [
--- a/backend/coreapp/platforms.py
+++ b/backend/coreapp/platforms.py
@@ -1,12 +1,11 @@
 import logging
 from dataclasses import dataclass, field
-from typing import Any, Dict, OrderedDict, Optional
+from typing import Any, Dict, OrderedDict
 from pathlib import Path
 import functools
 
 from coreapp.flags import COMMON_DIFF_FLAGS, COMMON_MIPS_DIFF_FLAGS, Flags
 from coreapp.models.preset import Preset
-from coreapp.models.profile import Profile
 from coreapp.models.scratch import Scratch
 from rest_framework.exceptions import APIException
 
--- a/backend/coreapp/platforms.py
+++ b/backend/coreapp/platforms.py
@@ -40,7 +40,7 @@ class Platform:
         return Scratch.objects.filter(platform=self.id).count()
 
     def to_json(
-        self, include_presets: bool = True, include_num_scratches: bool = False, profile: Optional[Profile] = None,
+        self, include_presets: bool = True, include_num_scratches: bool = False
     ) -> Dict[str, Any]:
         ret: Dict[str, Any] = {
             "id": self.id,
@@ -49,15 +49,10 @@ class Platform:
             "arch": self.arch,
             "has_decompiler": self.has_decompiler,
         }
-        if include_presets:       
-            # Here we are filtering the presets based on the owner
-            # if owner is None it means the preset is a system one => we keep it
-            # if owner is equal to profile it means it has been created by the user => we keep it
-            # otherwise we do not use it
+        if include_presets:
             ret["presets"] = [
                 PresetSerializer(p).data
                 for p in Preset.objects.filter(platform=self.id).order_by("name")
-                if p.owner is None or p.owner == profile
             ]
         if include_num_scratches:
             ret["num_scratches"] = self.get_num_scratches()
--- a/backend/coreapp/platforms.py
+++ b/backend/coreapp/platforms.py
@@ -1,11 +1,12 @@
 import logging
 from dataclasses import dataclass, field
-from typing import Any, Dict, OrderedDict
+from typing import Any, Dict, OrderedDict, Optional
 from pathlib import Path
 import functools
 
 from coreapp.flags import COMMON_DIFF_FLAGS, COMMON_MIPS_DIFF_FLAGS, Flags
 from coreapp.models.preset import Preset
+from coreapp.models.profile import Profile
 from coreapp.models.scratch import Scratch
 from rest_framework.exceptions import APIException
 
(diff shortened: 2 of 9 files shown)


backend/coreapp/migrations/0042_migrate_presets_to_db.py
```diff
@@ -823,6 +823,7 @@ def create_presets(apps: Apps, schema_editor: BaseDatabaseSchemaEditor) -> None:
 class Migration(migrations.Migration):
     dependencies = [
         ("coreapp", "0041_remove_project_and_github_stuff"),
+        ("coreapp", "0054_presets_owner"),
```
ethteck, line 826: is this intentional? I'm not a migration expert, but I don't think an older migration should depend on a newer one

backend/coreapp/platforms.py
```diff
@@ -48,10 +49,15 @@ def to_json(
...
-        if include_presets:
+        if include_presets:       
+            # Here we are filtering the presets based on the owner
+            # if owner is None it means the preset is a system one => we keep it
+            # if owner is equal to profile it means it has been created by the user => we keep it
+            # otherwise we do not use it
             ret["presets"] = [
                 PresetSerializer(p).data
                 for p in Preset.objects.filter(platform=self.id).order_by("name")
+                if p.owner is None or p.owner == profile
```
ethteck, line 60: Everyone should be able to see all presets. If I own project Banana and make a preset for it, I want everyone else on the site to be able to use that preset who's working on that  game. 

If there become too many presets to easily handle, we can make UI changes to make it easier to find the ones you want. but for now I think we don't need to filter like this. but I get your concern

backend/coreapp/views/compiler.py
```diff
@@ -34,11 +35,12 @@ def compilers_json() -> Dict[str, Dict[str, object]]:
     def platforms_json(
         include_presets: bool = True,
         include_num_scratches: bool = False,
+        profile: Optional[Profile] = None,
     ) -> Dict[str, Dict[str, object]]:
         ret: Dict[str, Dict[str, object]] = {}
 
         for platform in compilers.available_platforms():
-            ret[platform.id] = platform.to_json(include_presets, include_num_scratches)
+            ret[platform.id] = platform.to_json(include_presets, include_num_scratches, profile)
```
ethteck, line 43: likewise, I don't think we need any of this plumbing. the presets visible on the site shouldn't change based on who is fetching the data

backend/coreapp/views/preset.py
```diff
@@ -13,31 +15,48 @@
...
 
-class ReadOnly(BasePermission):
-    def has_permission(self, request: Any, view: Any) -> bool:
-        return request.method in SAFE_METHODS
+class IsOwnerOrReadOnly(BasePermission):
+    def has_object_permission(self, request: Any, view: Any, obj: Any) -> bool:
+        if request.method in SAFE_METHODS:
+            return True
+        if isinstance(obj, Preset):
+            return obj.owner == request.profile
```
ethteck, line 35: `or if request.profile.is_staff()`

if my syntax was wrong, we do this for making sure staff can delete scratches, so you should be able to use the same pattern

backend/coreapp/tests/test_preset.py
```diff
@@ -53,6 +72,54 @@ def test_create_preset_not_authenticated(self) -> None:
         except AssertionError:
             pass
 
+    def test_user_create_preset(self) -> None:
```
ethteck, line 75: these tests are great! could you add one to make sure anons can't make presets?

ethteck (changes requested):
awesome

backend/coreapp/migrations/0042_migrate_presets_to_db.py
axel7083: There is a problem with the migration, and I was not sure how to solve it...

The `0042_migrate_presets_to_db.py` is adding items to the Preset table, however they are using the latest `Preset` model class, which has the `owner` column, therefore leading to an error, when step 42 is triggered before step 54, which add the column...

I am not a migration expect either, and would appreciate some help on this particular issue!

backend/coreapp/views/preset.py
axel7083: This is inside the `IsOwnerOrReadonly` and the permissiosn are defined as followed

https://github.com/decompme/decomp.me/blob/9bc7f8a073c81582ddbcb1d03fc83e9ce9a4bed9/backend/coreapp/views/preset.py#L40

Therefore not requiring admin to be checked. I can add a unit test to ensure the expected behaviour if necessary

backend/coreapp/tests/test_preset.py
axel7083: There is already one which was already defined, I kept it.

https://github.com/decompme/decomp.me/blob/eca8cd2229974f7482ae77c0edff4a4786890d03/backend/coreapp/tests/test_preset.py#L66-L73

backend/coreapp/views/compiler.py
axel7083: I agree, and will be removing it when I will get some time!

backend/coreapp/platforms.py
axel7083: Good to know, I will restore previous code then

backend/coreapp/migrations/0042_migrate_presets_to_db.py
axel7083: This behaviour seems to have already been done in previous migration:

https://github.com/decompme/decomp.me/blob/6efceba4302d7064ec3d0614c5ed0bd12aed9424/backend/coreapp/migrations/0003_github_profile.py#L11

ethteck: I believe the way to handle this is the migration code shouldn't actually be using the real model classes, but I'm not 100% sure. we probably have screwed this up in the past. I'll see if I can take a look later today

backend/coreapp/views/preset.py
ethteck: ahhh missed that! thanks :)

backend/coreapp/views/compiler.py
axel7083: Done ✅

backend/coreapp/migrations/0042_migrate_presets_to_db.py
axel7083: Keep me updated

ethteck (approved):
other than the mypy issue, lgtm!

axel7083:
@ethteck now mypy should be happy :)

PR #1232, opened 2024-05-11
feat(frontend): allow presets to be explored in dedicated page

axel7083:
## Description

This PR make the presets easier to explore and integrate them as part of the UI with the related resources.

### `/preset`

The _new_ preset page allows to list all the presets by platform. I am reusing the component from the `NewScratchForm`.

image

## Related issues

Related to https://github.com/decompme/decomp.me/issues/648

### DIFF
--- a/frontend/src/app/(navfooter)/preset/page.tsx
+++ b/frontend/src/app/(navfooter)/preset/page.tsx
@@ -7,7 +7,7 @@ import { PresetList } from "@/components/PresetList"
 import * as api from "@/lib/api"
 import { get } from "@/lib/api/request"
 
-export default async function PresetPage() {
+export default async function Page() {
     const compilers = await get("/compiler")
 
     return (
@@ -17,7 +17,7 @@ export default async function PresetPage() {
     )
 }
 
-export function Presets({ serverCompilers }: {
+function Presets({ serverCompilers }: {
     serverCompilers: {
         platforms: {
             [id: string]: api.Platform
--- a/frontend/src/app/(navfooter)/preset/page.tsx
+++ b/frontend/src/app/(navfooter)/preset/page.tsx
@@ -1,10 +1,4 @@
-"use client"
-
-import React, { useState } from "react"
-
-import PlatformSelect from "@/components/PlatformSelect"
-import { PresetList } from "@/components/PresetList"
-import * as api from "@/lib/api"
+import { Presets } from "@/app/(navfooter)/preset/presets"
 import { get } from "@/lib/api/request"
 
 export default async function Page() {
@@ -14,38 +8,5 @@ export default async function Page() {
         <main className="mx-auto w-full max-w-3xl p-4">
             <Presets serverCompilers={compilers}/>
         </main>
-    )
-}
-
-function Presets({ serverCompilers }: {
-    serverCompilers: {
-        platforms: {
-            [id: string]: api.Platform
-        }
-        compilers: {
-            [id: string]: api.Compiler
-        }
-    }
-}): React.JSX.Element {
-
-    const platforms = Object.keys(serverCompilers.platforms)
-
-    const [platform, setPlatform] = useState<string>(platforms.length > 0?platforms[0]:"")
-
-    return (
-        <section>
-            <div>
-                <h2 className="pb-2 text-lg font-medium tracking-tight">Platforms</h2>
-                <PlatformSelect
-                    platforms={serverCompilers.platforms}
-                    value={platform}
-                    onChange={p => {
-                        setPlatform(p)
-                    }}
-                />
-            </div>
-            <h2 className="py-2 text-lg font-medium tracking-tight">Presets</h2>
-            <PresetList url={`/preset?platform=${platform}`}/>
-        </section>
     )
 }
--- a/frontend/src/app/(navfooter)/preset/page.tsx
+++ b/frontend/src/app/(navfooter)/preset/page.tsx
@@ -1,0 +1,51 @@
+"use client"
+
+import React, { useState } from "react"
+
+import PlatformSelect from "@/components/PlatformSelect"
+import { PresetList } from "@/components/PresetList"
+import * as api from "@/lib/api"
+import { get } from "@/lib/api/request"
+
+export default async function PresetPage() {
+    const compilers = await get("/compiler")
+
+    return (
+        <main className="mx-auto w-full max-w-3xl p-4">
+            <Presets serverCompilers={compilers}/>
+        </main>
+    )
+}
+
+export function Presets({ serverCompilers }: {
+    serverCompilers: {
+        platforms: {
+            [id: string]: api.Platform
+        }
+        compilers: {
+            [id: string]: api.Compiler
+        }
+    }
+}): React.JSX.Element {
+
+    const platforms = Object.keys(serverCompilers.platforms)
+
+    const [platform, setPlatform] = useState<string>(platforms.length > 0?platforms[0]:"")
+
+    return (
+        <section>
+            <div>
+                <h2 className="pb-2 text-lg font-medium tracking-tight">Platforms</h2>
+                <PlatformSelect
+                    platforms={serverCompilers.platforms}
+                    value={platform}
+                    onChange={p => {
+                        setPlatform(p)
+                    }}
+                />
+            </div>
+            <h2 className="py-2 text-lg font-medium tracking-tight">Presets</h2>
+            <PresetList url={`/preset?platform=${platform}`}/>
+        </section>
+    )
+}
--- a/frontend/src/components/PresetList.tsx
+++ b/frontend/src/components/PresetList.tsx
@@ -33,7 +33,7 @@ export function PresetList({ url, className, item, emptyButtonLabel }: Props): J
     const Item = item ?? PresetItem
 
     return (
-        <ul className={classNames("flex flex-col justify-center gap-[0.5em] overflow-hidden p-[1em] border-[1px] border-[solid] rounded-md border-gray-6 text-sm", className)}>
+        <ul className={classNames("flex flex-col justify-center gap-[0.5em] overflow-hidden rounded-md border-gray-6 text-sm", className)}>
             {results.map(preset => (
                 <Item hideIcon key={preset.id} preset={preset} />
             ))}
--- a/frontend/src/components/PresetList.tsx
+++ b/frontend/src/components/PresetList.tsx
@@ -1,0 +1,74 @@
+"use client"
+
+import React, { ReactNode } from "react"
+
+import Link from "next/link"
+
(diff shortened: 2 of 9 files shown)


conorgolden1:
I am looking to display comments on the user profile page. So we might be running into design conflicts here with the changes to the user page. Also I am wondering what would happen in your implementation you requested to show more of the users scratches? Would that just extend the preset section further down the page? 

I am just thinking maybe to future proof this a little we add tabs to the user profile page that can then be added onto with future features such as this.

ethteck:
I'm not sure comments belong on a user page since you lose the context of the scratch to just have a bunch of comments. or maybe there's a way to do that you have in your head I'm not considering. but yeah I am down with the tabs idea, I think

ethteck:
Another thought: 

I think showing a few presets at the top of a platform page is good, but if we list all of them, the scratches will be quite far down the page. maybe we could do the tabs thing there too, or a two-column view? not sure..

axel7083:
> Another thought:
> 
> I think showing a few presets at the top of a platform page is good, but if we list all of them, the scratches will be quite far down the page. maybe we could do the tabs thing there too, or a two-column view? not sure..

I simplified this PR, by only adding the `/preset` page for now.

I also think it would be great to have some presets in the platform page, but adding a tab section could be another PR!

ethteck:
this looks great! a few thoughts:

1. it'd be cool if we had a sortable PresetList (number of scratches, recent activity, alphabetical, creation date, all ascending or descending). we can address this in a future PR, since it'd require backend changes
2. It'd also be cool if we could show the number of scratches in the Preset item somewhere. Exposing this in a nice way might warrant changing the backend, so we can tackle this later too.
3. We can probably do away with the platform logo on this particular page, since we're filtering by platform anyway.

ethteck:
4. the platform logo in the filter selection at the top of the page should not be a link

(3 and 4 should probably be addressed in this PR before merge, and I can write an issue for 1 and 2)

axel7083:
> 4. the platform logo in the filter selection at the top of the page should not be a link
> 
> (3 and 4 should probably be addressed in this PR before merge, and I can write an issue for 1 and 2)

@ethteck 3 ans 4 have been implemented

ethteck (approved):

frontend/src/app/(navfooter)/preset/page.tsx
```diff
@@ -0,0 +1,51 @@
+"use client"
```
bates64, line 1: Fully client rendered pages should be avoided :-1:

frontend/src/app/(navfooter)/preset/page.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import React, { useState } from "react"
```
bates64, line 3: Why are you importing `React`? This prevents tree-shaking
There's no need to type things `React.JSX.Element`; see the rest of the codebase

frontend/src/app/(navfooter)/preset/page.tsx
```diff
@@ -0,0 +1,51 @@
...
+    serverCompilers: {
+        platforms: {
+            [id: string]: api.Platform
+        }
+        compilers: {
+            [id: string]: api.Compiler
+        }
+    }
+}): React.JSX.Element {
+
```
bates64, line 30:
```suggestion
```

frontend/src/app/(navfooter)/preset/page.tsx
```diff
@@ -0,0 +1,51 @@
...
+    const platforms = Object.keys(serverCompilers.platforms)
+
+    const [platform, setPlatform] = useState<string>(platforms.length > 0?platforms[0]:"")
```
bates64, line 33:
```suggestion
    const [platform, setPlatform] = useState<string>(platforms.length > 0 ? platforms[0] : "")
```

frontend/src/app/(navfooter)/preset/page.tsx
```diff
@@ -0,0 +1,51 @@
...
+    return (
+        <section>
+            <div>
+                <h2 className="pb-2 text-lg font-medium tracking-tight">Platforms</h2>
+                <PlatformSelect
+                    platforms={serverCompilers.platforms}
+                    value={platform}
+                    onChange={p => {
+                        setPlatform(p)
+                    }}
```
bates64, line 44: I'd just do
```suggestion
                    onChange={setPlatform}
```

frontend/src/app/(navfooter)/preset/page.tsx
```diff
@@ -0,0 +1,51 @@
...
+    }
+}): React.JSX.Element {
+
+    const platforms = Object.keys(serverCompilers.platforms)
+
+    const [platform, setPlatform] = useState<string>(platforms.length > 0?platforms[0]:"")
+
+    return (
+        <section>
+            <div>
```
bates64, line 37: What's the point of this div? It has no styling nor does it have any semantic meaning

frontend/src/components/PresetList.tsx
```diff
@@ -0,0 +1,76 @@
+"use client"
+
+import React, { ReactNode } from "react"
```
bates64, line 3: Ditto, don't import all of React, and don't annotate component functions with a return type, there's no need

frontend/src/components/PresetList.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+import Link from "next/link"
+
+import classNames from "classnames"
+
+import AsyncButton from "@/components/AsyncButton"
+import Button from "@/components/Button"
+import LoadingSpinner from "@/components/loading.svg"
+import { PlatformIcon } from "@/components/PlatformSelect/PlatformIcon"
+import styles from "@/components/ScratchList.module.scss"
```
bates64, line 13: Aah, let's not import styles from another component like this.

Much better would be to port the styles into tailwind, and/or figure out how to abstract these into components so they can be reused properly.

The goal long-term is to stop using SCSS modules in favour of Tailwind and components.

frontend/src/components/PresetList.tsx
```diff
@@ -0,0 +1,76 @@
...
+import * as api from "@/lib/api"
```
bates64, line 14: Ideally we don't introduce new code that does this, use more granular imports

frontend/src/components/PresetList.tsx
```diff
@@ -0,0 +1,76 @@
...
+export function PresetItem({ preset, hideIcon }: { preset: api.Preset, hideIcon?: boolean }): React.JSX.Element {
+    const compilersTranslation = useTranslation("compilers")
+    const compilerName = compilersTranslation.t(preset.compiler)
+
+    return (
+        <div className="rounded-md border border-gray-6 p-[1em] text-sm">
+            <div className="flex items-center gap-2">
+                {
+                    hideIcon ? <></>:<PlatformIcon platform={preset.platform} className="w-[1.2em]"/>
+                }
```
bates64, line 68:
```suggestion
                {hideIcon && <PlatformIcon platform={preset.platform} className="w-[1.2em]"/>}
```

frontend/src/components/PresetList.tsx
```diff
@@ -0,0 +1,76 @@
...
+export function PresetList({ url, className, item, emptyButtonLabel }: Props): React.JSX.Element {
+    const { results, isLoading, hasNext, loadNext } = api.usePaginated<api.Preset>(url || "/preset")
+    if (results.length === 0 && isLoading) {
+        return <div className={classNames(styles.loading, className)}>
+            <LoadingSpinner width="1.5em" height="1.5em" />
+            Just a moment...
+        </div>
+    }
+
+    const Item = item || PresetItem
```
bates64, line 34:
```suggestion
    const Item = item ?? PresetItem
```

bates64 (changes requested):

bates64:
I like this PR! Thanks for the contribution - just some small code style things to fix up.

frontend/src/components/PresetList.tsx
bates64: If you like this can be a TODO if you make an issue

frontend/src/components/PresetList.tsx
axel7083: Done ✅

frontend/src/components/PresetList.tsx
axel7083: Done ✅

frontend/src/components/PresetList.tsx
axel7083: I change to granular import instead of wildcard

frontend/src/components/PresetList.tsx
axel7083: I change to tailwindcss classes ✅

frontend/src/components/PresetList.tsx
axel7083: I always have some preference to precise the return type for readability, but okey removed

frontend/src/app/(navfooter)/preset/page.tsx
axel7083: removed

frontend/src/app/(navfooter)/preset/page.tsx
axel7083: applied

frontend/src/app/(navfooter)/preset/page.tsx
axel7083: applied

frontend/src/app/(navfooter)/preset/page.tsx
axel7083: I changed to `import { type JSX } from 'react'` when I could, otherwise removed it

frontend/src/app/(navfooter)/preset/page.tsx
axel7083: Removed

axel7083:
Result can be visible through the vercel deployment https://frontend-git-fork-axel7083-feature-frontend-presets-decompme.vercel.app/preset

frontend/src/app/(navfooter)/preset/page.tsx
```diff
@@ -0,0 +1,12 @@
+import { Presets } from "@/app/(navfooter)/preset/presets"
```
bates64, line 1: Perhaps better to do ./presets.tsx in future but this is fine

bates64 (approved):

bates64:
Thanks for the fixes :) Looks like there are some conflicts. Soon as they're resolved we can merge this👍

axel7083:
> Thanks for the fixes :) Looks like there are some conflicts. Soon as they're resolved we can merge this👍

Yeahhh, should be good now :)

axel7083:
@bates64 ready to be merge !

PR #1236, opened 2024-05-13
Scroll to line when clicking line number

1superchip:
Scroll to line in source tab when line numbers in diff are clicked.

This does not highlight the newly selected line.

Related to https://github.com/decompme/decomp.me/issues/115.
"Click a line number to jump to it in the source"

This should be reviewed.

### DIFF
--- a/frontend/src/components/Diff/Diff.tsx
+++ b/frontend/src/components/Diff/Diff.tsx
@@ -94,7 +94,7 @@ function scrollToSourceFromLineNumber(cell: api.DiffCell | undefined) {
         const lineNumbersEle = document.getElementsByClassName("cm-gutter cm-lineNumbers").item(0)
         const scrollerEle = document.getElementsByClassName("cm-scroller").item(0)
         if (lineNumbersEle) {
-            const lineNumberEle = lineNumbersEle.children[cell.src_line]
+            const lineNumberEle = lineNumbersEle.children[cell.src_line] as HTMLElement
             if (lineNumberEle) {
                 scrollerEle.scroll({
                     left: lineNumberEle.offsetLeft,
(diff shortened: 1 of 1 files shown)


ethteck:
I forgot to write here, but I think we should adjust this to be more React friendly before merging. It's good to have this as a starting point, though. but stuff like `getElementsByClassName` isn't ideal for a react project like this

bates64:
Agreed with @ethteck. The way this PR is currently written requires that there only be one CodeMirror on the page at once, which is the case today, but might not be the case in future.

To do this in a more React-like way, a [ref](https://react.dev/learn/referencing-values-with-refs) to the CodeMirror element should be provided via [context](https://react.dev/learn/passing-data-deeply-with-context), probably from the top-level Scratch editor component. Then when this button is clicked, you can do `.getElementsByClassName` on the ref instead of on the entire document.

mkst:
Would the context be a reference to the CodeMirror element itself? from that tutorial on context:

```js
import { LevelContext } from './LevelContext.js';

export default function Section({ level, children }) {
  return (
    <section className="section">
      <LevelContext.Provider value={level}>  /* what would we be wrapping in a context, what what would the value be? */
        {children}
      </LevelContext.Provider>
    </section>
  );
}
```

in `DecompilePanel.tsx` we do this:
```
            {typeof decompiledCode == "string" && <>
                <CodeMirror
                    className={styles.editor}
                    value={decompiledCode}
                    valueVersion={valueVersion}
                    viewRef={viewRef}
                    extensions={[
                        decompileSetup,
                        cpp(),
                        compareExtension,
                    ]}
                />
            </>}
```
would this be wrapped in a `CodeMirrorContext.Provider value={what?}` or do we do `useRef()` inside CodeMirror and wrap a context around the 
```js
    return <div
        ref={el}
        onMouseMove={debouncedOnMouseMove}
        className={classNames(styles.container, className)}
        style={{
            "--cm-font-size": `${fontSize}px`,
            "--cm-container-width": `${width}px`,
        } as CSSProperties}
    />
```

mkst:
Gonna close this one as the change has been implemented - heavily based on your code, so thank you very much!

PR #1237, opened 2024-05-13
Fixed user page sorting

### LINKED ISSUE #1234: Make user page scratchlist sortable

conorgolden1:
fixes #1234 

[Screencast from 2024-05-13 14-04-48.webm](https://github.com/decompme/decomp.me/assets/151766738/ca865c6a-a82d-4bc7-81e5-33a243dfe662)

### DIFF
--- a/backend/coreapp/views/scratch.py
+++ b/backend/coreapp/views/scratch.py
@@ -303,7 +303,7 @@ class ScratchViewSet(
         filters.OrderingFilter,
     ]
     search_fields = ["name", "diff_label"]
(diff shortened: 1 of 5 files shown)


ethteck:
LGTM! I think we should have most/least matched rather than only the current option (which appears to be ascending) - this can apply to all the sortable lists

conorgolden1:
> LGTM! I think we should have most/least matched rather than only the current option (which appears to be ascending) - this can apply to all the sortable lists

Should be all set with the new changes

PR #1244, opened 2024-05-18
Add more MWCCPSP compilers

Xeeynamo:
Add more PlayStation Portable compilers from Metrowerks from 2005 and 2006.


mkst:
I feel like this PR stomped over mine?

mkst:
Or actually looks like it's exactly the same change 🤣

mkst:
Great minds and all that 😉

mkst:
For reference this is the associated commit in the compilers repo: https://github.com/decompme/compilers/commit/de496e580011032bab6fb984d3200ad50d438f7a

PR #1247, opened 2024-05-19
Remove old preset-creation migration from the nation

ethteck:
This migration doesn't actually do anything other than migrate some data for the live site, so it's not worth keeping around if it's going to cause issues


ethteck:
oh, nevermind, this breaks the tests.

PR #1258, opened 2024-05-26
chore(frontend): adding dependabot

axel7083:
Some package in the `frontend/packages.json` are a bit outdated, keeping them up to date manually can be often tedious.

Github has the `dependabot` allowing to automate this process.

See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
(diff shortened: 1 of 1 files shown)


bates64:
I'm against this until we have tests. We've had issues in the past updating minor versions of packages that have introduced subtle bugs. Just updating a package for the sake of keeping it up-to-date has no utility.

PR #1270, opened 2024-06-04
Use modified PS2 binutils for PSP

sozud:
SOTN uses a Metrowerks compiler that seems to emit the same sort of debug info as the PS2 compilers. This commit changes the binutils to use the modified PS2 binutils. From what I see locally it seems to work fine.

image

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -85,8 +85,10 @@ RUN if [ "${ENABLE_MSDOS_SUPPORT}" = "YES" ] || \
     wine; \
     fi
 
+ARG ENABLE_PSP_SUPPORT
+
 # Patched mips binutils
-RUN if [ "${ENABLE_PS2_SUPPORT}" = "YES" ]; then \
+RUN if [ "${ENABLE_PS2_SUPPORT}" = "YES" ] || [ "${ENABLE_PSP_SUPPORT}" = "YES" ]; then \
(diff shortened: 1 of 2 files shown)


mkst:
Could you update the Dockerfile too? Move the 
`ARG ENABLE_PSP_SUPPORT` just before line 89 and add the condition to the `if` statement?
https://github.com/decompme/decomp.me/blob/7cb11c2c6bfb69639305cd83030e49225ade11fe/backend/Dockerfile#L89

sozud:
My docker is messed up locally but it looks like it worked in CI

```
#16 [build  7/13] RUN if [ "YES" = "YES" ] || [ "${ENABLE_PSP_SUPPORT}" = "YES" ]; then     wget "https://github.com/decompals/binutils-mips-ps2-decompals/releases/download/v0.4/binutils-mips-ps2-decompals-linux-x86-64.tar.gz" &&     tar xvzf binutils-mips-ps2-decompals-linux-x86-64.tar.gz -C /usr/bin mips-ps2-decompals-as mips-ps2-decompals-nm mips-ps2-decompals-objdump &&     rm binutils-mips-ps2-decompals-linux-x86-64.tar.gz &&     chmod +x /usr/bin/mips-ps2-decompals-*;     fi
#16 0.195 --2024-06-04 05:28:19--  https://github.com/decompals/binutils-mips-ps2-decompals/releases/download/v0.4/binutils-mips-ps2-decompals-linux-x86-64.tar.gz
```

The Vercel error seems unrelated? I can't redeploy it.

PR #1273, opened 2024-06-09
fix: profile field serialization in compiler view

axel7083:
## Description

When creating custom presets, the `/api/compiler` is including the presets for each platforms, however when serialiazing the owner field a KeyError is raised.

As visible in the code bellow, we are tyring to get the `request` from the context.

https://github.com/decompme/decomp.me/blob/294a2dc1b9fe1d6a033e07d88c92bf4a885d4769/backend/coreapp/serializers.py#L56-L58

However, in this specific context (serialiazing the owner field of a custom preset we include for a platform), it is not available. 
image

We can clearly notice however that the `serialize_profile` function is not using the request argument we are sending, it uses the `profile` only. Therefore an easy fix is to remove the request.

## Tests

- [x] regression unit tests has been added

### DIFF
--- a/backend/coreapp/tests/test_preset.py
+++ b/backend/coreapp/tests/test_preset.py
@@ -77,6 +77,19 @@ class PresetTests(BaseTestCase):
         preset = self.create_preset(DUMMY_PRESET_DICT)
         assert preset.owner is not None
         assert preset.owner.pk == self.user.pk
+        
+    def test_list_compiler_with_presets(self) -> None:
+        user = self.create_user()
+        self.create_preset(DUMMY_PRESET_DICT)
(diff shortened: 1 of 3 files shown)


backend/coreapp/tests/test_preset.py
```diff
@@ -78,6 +78,22 @@ def test_user_create_preset(self) -> None:
         assert preset.owner is not None
         assert preset.owner.pk == self.user.pk
 
+    def test_list_compiler_with_presets(self) -> None:
```
axel7083, line 81:
```suggestion
    def test_list_compiler_with_custom_presets(self) -> None:
```

PR #1274, opened 2024-06-10
Add mwcc_43_202 compiler (Wii MW 1.6)

encounter:
This already exists in [compilers](https://github.com/decompme/compilers/tree/main/platforms/gc_wii/mwcc_43_202), but appears to have been missed here.


mkst:
Thanks!

PR #1275, opened 2024-06-10
Family fix!

ethteck:


backend/coreapp/views/scratch.py
```diff
@@ -517,14 +530,21 @@ def export(self, request: Request, pk: str) -> HttpResponse:
...
             ).order_by("creation_time")
+        elif (
+            scratch.target_assembly.elf_object is not None
+            and len(scratch.target_assembly.elf_object) > 0
+        ):
+            family = Scratch.objects.filter(
+                target_assembly__hash=scratch.target_assembly.hash,
+            ).order_by("creation_time")
+        else:
+            family = Scratch.objects.filter(slug=scratch.slug)
```
simonlindholm, line 547: deduplicate this and create a function that returns `family`? (letting this caller add an .order_by call)

backend/coreapp/views/scratch.py
```diff
@@ -169,23 +169,36 @@ def scratch_etag(request: Request, pk: Optional[str] = None) -> Optional[str]:
...
+            scratch.target_assembly.elf_object is not None
+            and len(scratch.target_assembly.elf_object) > 0
+        ):
+            family = Scratch.objects.filter(
+                target_assembly__hash=scratch.target_assembly.hash,
+            )
+        else:
+            family = Scratch.objects.filter(slug=scratch.slug)
 
         return str(hash((family, request.headers.get("Accept"))))
```
simonlindholm, line 203: It's worth double-checking + adding a test that this actually works as it's supposed to. I wouldn't be surprised if hash() on a query set returns its id(), i.e. pretty much a random number, and that you need to call list() on it or even further serialize it for it to work.

BTW it's also not great that this query is performed twice; once for the etag and once for the content. Maybe it would be faster just to remove the etag.

PR #1289, opened 2024-06-29
Add PS2 compiler ee-gcc29-991111b/r4

TheOnlyZac:
This is the compiler for Sly Cooper on PS2.


backend/coreapp/compilers.py
```diff
@@ -538,6 +538,12 @@ def available_platforms() -> List[Platform]:
     cc='${COMPILER_DIR}/bin/ee-gcc -c $COMPILER_FLAGS "$INPUT" -o "$OUTPUT"',
 )
 
+EE_GCC29_99111B_R4 = GCCPS2Compiler(
```
Yotona, line 541: Did you mean 991111B with four 1s?

backend/coreapp/compilers.py
```diff
@@ -1519,6 +1525,7 @@ def available_platforms() -> List[Platform]:
     EE_GCC29_990721,
     EE_GCC29_991111,
     EE_GCC29_991111A,
+    EE_GCC29_99111B_R4,
```
Yotona, line 1528: Here too, probably meant with four 1s

backend/coreapp/compilers.py
TheOnlyZac: Yes, thank you! Fixed both.

TheOnlyZac:
Might need to rerun the checks since my PR on the compilers repo was merged.

ethteck:
@TheOnlyZac this failed https://github.com/decompme/decomp.me/actions/runs/9727725876/job/27109038662?pr=1289

Are you sure you named the output file correctly? I saw in the PR, the archive name doesn't match the compiler ID. it doesn't include the b4 part.. https://github.com/decompme/compilers/pull/15/files#diff-8377b3e3740a3fcd9f682e5fb55425f2fdbece1791854b9e5013e7f1a5e60e7eR224 not sure if that's the issue, but just saw it

TheOnlyZac:
Haven't had a chance to look at this but I will soon.

backend/coreapp/compilers.py
```diff
@@ -538,6 +538,12 @@ def available_platforms() -> List[Platform]:
     cc='${COMPILER_DIR}/bin/ee-gcc -c $COMPILER_FLAGS "$INPUT" -o "$OUTPUT"',
 )
 
+EE_GCC29_991111B_R4 = GCCPS2Compiler(
+    id="ee-gcc2.9-991111b-r4",
+    platform=PS2,
+    cc='${COMPILER_DIR}/bin/ee-gcc -c $COMPILER_FLAGS "$INPUT" -o "$OUTPUT"',
```
mkst, line 544: this is a Windows executable `ee-gcc.exe`, so would need to be invoked with wine or wibo.
```
$file ee-gcc.exe
ee-gcc.exe: PE32 executable (console) Intel 80386, for MS Windows
```

mkst:
Gonna close this PR as it's gone stale. Feel free to re-open once you have the compiler working.

Mc-muffin:
A bit of an addendum to this, while afaik Sly Cooper is using SN GCC 2.95.X now and it was our understanding that 991111b/r4 was just a windows build of vanilla 991111 turns out there's at least 1 known codegen difference in [AIFF_GetInfo](https://decomp.me/scratch/T2H4L) which causes a slightly different scheduling and stack reservation.
While investigating why I eventually found out that 991111b/r4 is indeed just 991111 but with Dylan Cuthbert's VU0 extensions (which make the compiler aware of the extended VU0 register set for scheduling purposes), with that in mind I compiled an equivalent Linux version which (if eventually needed) could be added to the site instead of the Windows version
[ee-gcc-991111-b4.tar.gz](https://github.com/user-attachments/files/28166478/ee-gcc-991111-b4.tar.gz)

PR #1306, opened 2024-07-30
Add msvc8.0 win32

widberg:
Goes with https://github.com/decompme/compilers/pull/22

### DIFF
--- a/backend/compilers/compilers.linux.yaml
+++ b/backend/compilers/compilers.linux.yaml
@@ -198,3 +198,4 @@ win32:
   - msvc6.6
   - msvc7.0
   - msvc7.1
+  - msvc8.0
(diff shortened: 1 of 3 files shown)


mkst:
Currently failing CI with:
```
coreapp.error.CompilationError: Compiler error: cl : Command line error D8037 : cannot create temporary il file; clean temp directory of old il files
```
I dunno what an `il` file is or why it can't be created. is there a flag to tell `cl.exe` the path to put `il` files?

widberg:
Thank you for merging this. I find it odd that I didn't run into that issue locally in my Ubuntu 22.04.4 VM with wine from the default Canonical apt repositories (wine 6.0.3~repack-1). Sorry you had to debug that, but thank you for doing so.

mkst:
It's very strange, the issue only occurred when the compiler was being ran inside Docker 🤷

PR #1327, opened 2024-08-16
Win32 Platform tweaks

jpires:
Hello guys, 

I made a couple of changes that I think can be useful for the rest of the community.

### X86 Assembly Syntax choice
By default, the win32 platform displays assembly in the AT&T syntax.
However, most of the tools that I use, Ghidra and x64dbg show assembly in the Intel syntax.

I have added a flag on the Diff section for Win32 to able able to select the preferred syntax for a scratch.

### Replace GNU AS with NASM
The reason to do this is that GNU AS doesn't support the name mangling that the MSVC uses, so it makes it impossible to match C++ names, as they cannot be in the original assembly code.
NASM doesn't have this limitation and allows us to use the MSVC mangled names.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -54,6 +54,7 @@ RUN apt-get -y update && apt-get install -y \
     libnl-route-3-200 \
     libprotobuf-dev \
     libtinfo5 \
+    nasm \
     netcat \
     software-properties-common \
     unzip \
--- a/backend/coreapp/diff_wrapper.py
+++ b/backend/coreapp/diff_wrapper.py
@@ -133,7 +133,7 @@ class DiffWrapper:
 
     @staticmethod
     def parse_objdump_flags(diff_flags: List[str]) -> List[str]:
-        known_objdump_flags = ["-Mno-aliases"]
+        known_objdump_flags = ["-Mno-aliases", "-Mintel", "-Matt"]
         known_objdump_flag_prefixes = ["-Mreg-names=", "--disassemble="]
         ret = []
 
--- a/backend/coreapp/flags.py
+++ b/backend/coreapp/flags.py
@@ -323,3 +323,10 @@ COMMON_WATCOM_FLAGS: Flags = [
     Checkbox("watcom_signedchar", "-j"),
     Checkbox("watcom_fpu", "-fpi87"),
 ]
+
+COMMON_X86_DIFF_FLAGS: Flags = [
+    FlagSet(
+        id="x86_asm_syntax",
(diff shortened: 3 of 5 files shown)


ethteck:
Hi, thanks for the PR! Are you in the discord? It might be easier to discuss the implications of it there, but to echo what I said:

I'm curious if you've tried this locally. the only reason we weren't using intel syntax is because asm_differ's x86 support was done with att syntax in mind. we have an outstanding issue to add support for intel syntax, but I think only some preliminary work was done on that so far: https://github.com/simonlindholm/asm-differ/issues/123

Changing to an entirely different assembler also worries me a bit, because I'd hate to break existing workflows / scratches

jpires:
Hi, no, I am not on Discord, but I will try to join in the next couple of days.

I didn't fully try this patch before submitting it, but I run the individual changes locally and was able to compile and match a few scratches.
The only thing that I didn't test, and it doesn't seem to be working at is the assembly syntax choice, it doesn't use the default choice. e.g. it doesn't start by showing the Intel syntax. To be honest, I only stumble on this solution the night before and my quick test seemed to work.

Anyway, the matches that I tested don't seem to be an issue for asm_differ, it matches them correctly.
To be honest they are not that big, so maybe there are some corner cases that I haven't hit.

You are right, about the new assembler, I hadn't even thought about that, but yes, it has at least a couple of changes to the GNU as, that may break existing workflows.
From my testing, it doesn't break old scratches, but it does change the assembly that is accepted.
So, if you try to recreate a scratch by copying the assembly, it may not work anymore.

1superchip:
Checking the target assembly of these scratches will show how I’ve created assembly that works with GNU AS

Quoting symbols and setting some other name to the mangled symbol to use in instructions like `call` works

https://decomp.me/scratch/t41B2 has a mangled symbol with templates in the target assembly

https://decomp.me/scratch/MTd9z
Target Assembly:
```
.set lit_0042FABC, "?Crc_Update@@YAKKPBXH@Z"
/*  0042FABC: E8 5F FF FF FF     */ call        lit_0042FABC
```

https://decomp.me/scratch/VTEUi
Target Assembly:
```
"?Replay_Deinit@@YAXXZ":
/*  0040F870: C3                 */ ret
```

jpires:
@1superchip thanks for those links
That does solve my problem with the name mangling.

ethteck:
@jpires does this mean we can close this PR, or is there more that's not covered in the above comment?

jpires:
I am fine now.
I will close it.

PR #1329, opened 2024-08-21
Experimental objdiff integration (client-side)

encounter:
Behind an experimental toggle in Editor settings.

See https://github.com/encounter/objdiff/tree/main/objdiff-wasm/src for the library implmentation. It spins up a web worker that runs the objdiff wasm binary, and uses protobuf for fast result deserialization.

This implements a new Diff view that can handle objdiff's schema, and has various performance improvements around React re-rendering.

Overall, the changes are quite noninvasive, which is good :slightly_smiling_face: 

Caveats:
- Currently requires the scratch's diff label to exist as a symbol in both the target and current objects. I'm pondering how to implement symbol selection.
- Untested on platforms other than GC/Wii, though it should support x86/COFF and ARM (GBA, DS, 3DS) as well. No MIPS until we have a :crab: disassembler.
- Need to implement asm-differ-like scoring.


ethteck (approved):
I am not qualified to review the frontend shenanigans, but I'm happy with this being merged in for now and we can slowly make it more and more of a thing until it becomes the default

PR #1344, opened 2024-09-14
Friendly hovertext for TimeAgo

### LINKED ISSUE #1331: Hover-over timestamps are unfriendly
image

It's good that we show the day and time. I think we can avoid showing a resolution beyond single seconds, though. and adding some spaces for readability would be ideal.

maybe just `2024-07-28 18:40:31 +09:00`

We should check to see if this is an issue in other places besides just the about scratch page, for good measure

mkst:
Closes #1331.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -46,6 +46,7 @@
         "react": "^18.2.0",
         "react-colorful": "^5.6.1",
         "react-contenteditable": "^3.3.7",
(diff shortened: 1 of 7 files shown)


mkst:
I'm guessing the date parsing is happening clientside, which is good for timezones (it's 10:30 in the UK as i write this):

image

PR #1370, opened 2024-10-27
Try to fix the overabundance of profiles

ethteck:
and backend dep updates


backend/coreapp/middleware.py
```diff
@@ -51,6 +51,7 @@ def middleware(request: Request) -> Response:
             "node-fetch" in request.headers["User-Agent"]
```
mkst, line 51: Assuming user agent is a string, can this check be removed as it'll be caught by `"node" in ...`?

ethteck: I thought it wasn't a string, but yeah you're right

mkst (approved):

PR #1394, opened 2024-12-08
Tailwindify

ethteck:
Don't merge until #1392


frontend/src/app/(navfooter)/new/NewScratchForm.tsx
```diff
@@ -18,7 +18,28 @@ import basicSetup from "@/lib/codemirror/basic-setup";
...
 import getTranslation from "@/lib/i18n/translate";
 
-import styles from "./new.module.scss";
+interface FormLabelProps {
+    children: React.ReactNode;
+    htmlFor?: string;
+    small?: string;
+}
+
+function FormLabel({ children, htmlFor, small }: FormLabelProps) {
```
bates64, line 27: Call this `htmlEl` (i.e. element) rather than `htmlFor` because that has a very specific other meaning: https://developer.mozilla.org/en-US/docs/Web/API/HTMLLabelElement/htmlFor

```suggestion
function FormLabel({ children, htmlEl, small }: FormLabelProps) {
```

frontend/src/app/(navfooter)/new/NewScratchForm.tsx
```diff
@@ -18,7 +18,28 @@ import basicSetup from "@/lib/codemirror/basic-setup";
...
+    const Tag = htmlFor ? "label" : "p";
+    return (
+        <Tag
+            className="m-0 block p-2.5 font-semibold text-[0.9em] text-[color:var(--g1700)]"
+            {...(htmlFor && { htmlFor })}
```
bates64, line 32: prefer ?? probably

frontend/src/app/(navfooter)/new/NewScratchForm.tsx
```diff
@@ -272,26 +293,28 @@ export default function NewScratchForm({
...
+                            className="w-full"
                             options={compilerChoiceOptions}
                             value={compilerId}
                             onChange={setCompiler}
                         />
                     </div>
-                    <div className={styles.compilerChoiceOr}>or</div>
-                    <div>
-                        <span className={styles.compilerChoiceHeading}>
+                    <div className="flex-0 px-2 py-2 text-center text-[0.8rem] text-[color:var(--g500)] min-[400px]:px-4">
```
bates64, line 309: Surprised we don't have a tailwind config value for this: `text-[color:var(--g500)] `, like `text-grey-500` or something. If not we should raise an issue about doing that

bates64 (changes requested):

frontend/src/app/(navfooter)/new/NewScratchForm.tsx
mkst: There are still a lot of places where color:var is used:

```
mark@carbon:~/github/decomp.me/frontend$ grep -RlE "color:\s?var" src | sort
src/app/globals.scss
src/app/(navfooter)/new/NewScratchForm.tsx
src/components/AsyncButton.module.scss
src/components/Breadcrumbs.module.scss
src/components/Button.module.scss
src/components/ColorSchemePicker.module.scss
src/components/compiler/CompilerOpts.module.css
src/components/Diff/Diff.module.scss
src/components/Editor/CodeMirror.module.scss
src/components/Nav/Search.module.scss
src/components/Nav/UserMenuItems.module.scss
src/components/PlatformSelect/PlatformSelect.module.scss
src/components/ScoreBadge.module.scss
src/components/ScratchList.module.scss
src/components/Scratch/panels/AboutPanel.module.scss
src/components/Scratch/panels/DecompilePanel.module.scss
src/components/Scratch/ScratchToolbar.module.scss
src/components/Tabs.module.scss
src/components/ThemePicker.module.scss
src/components/VerticalMenu.module.scss
```

There doesn't appear to be a 1:1 mapping of the tailwind `X-gray-N` for these `--gNNN` values (or maybe there is and i'm dumb?)...

but perhaps this could be ripped out in a future PR?

PR #1399, opened 2024-12-16
Initial support for Pascal syntax

LLONSIT:
- This PR adds minimal support for the Pascal language to mostly highlight some keywords and types.
- Note that this was mostly implemented to support SGI Pascal (the Pascal extension that IDO uses) and not common modern dialects like Free Pascal or Delphi.

### DIFF
--- a/frontend/src/lib/codemirror/pascal.js
+++ b/frontend/src/lib/codemirror/pascal.js
@@ -24,19 +24,19 @@ var isOperatorChar = /[+\-^&%=<>!?|\/]/;
 function tokenBase(stream, state) {
   var ch = stream.next();
 
-  if (ch == "#" && state.startOfLine) {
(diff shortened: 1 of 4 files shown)


ethteck (approved):

frontend/src/lib/codemirror/pascal.js
```diff
@@ -0,0 +1,160 @@
+/* Based on @codemirror/laegacy-modes/pascasl */
```
bates64, line 1:
```suggestion
/* Based on @codemirror/legacy-modes/pascal */
```

bates64 (changes requested):
Otherwise, LGTM

frontend/src/lib/codemirror/pascal.js
LLONSIT: Oops

PR #1419, opened 2025-01-27
[WIP] Add Borland 3.1 compiler to MS-DOS platform

Gillou68310:
Depends on
https://github.com/simonlindholm/asm-differ/pull/178
https://github.com/decompme/compilers/pull/28

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,9 @@ jobs:
       - name: Install msdos assembler
         run: |-
           wget https://github.com/OmniBlade/binutils-gdb/releases/download/omf-build/omftools.tar.gz
-          sudo tar xvzf omftools.tar.gz -C /usr/bin jwasm omf-nm omf-objdump
+          sudo tar xvzf omftools.tar.gz -C /usr/bin jwasm
+          wget https://github.com/sillyc0n/binutils-gdb/releases/download/2.42-build7/omftools-2.42-build7-linux-x86_64.tar.gz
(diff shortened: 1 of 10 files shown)


backend/coreapp/compilers.py
```diff
@@ -1472,6 +1479,20 @@ def available_platforms() -> List[Platform]:
     cc=WATCOM_CXX,
 )
 
+BORLAND_MSDOS_CC = (
+    'cat "$INPUT" | unix2dos > dos_src.c && '
+    "echo \"\$_hdimage = '+0 ${COMPILER_DIR} +1'\" > .dosemurc && "
+    '(HOME="." /usr/bin/dosemu -quiet -dumb -f .dosemurc -K . -E "D:\\bin\\bcc.exe -ID:\\include ${COMPILER_FLAGS} -c -oout.o dos_src.c") && '
+    "rm .dosemurc && "
```
mkst, line 1486: Don't need to worry about the rm, the execution is done in a sandbox

mkst:
You'll need to update the asm-differ dependency in the [poetry.lock](https://github.com/decompme/decomp.me/blob/main/backend/poetry.lock#L54) once your PR has been merged there, but other than that, this looks good to go!

.github/workflows/ci.yml
```diff
@@ -59,6 +59,8 @@ jobs:
         run: |-
           wget https://github.com/OmniBlade/binutils-gdb/releases/download/omf-build/omftools.tar.gz
```
mkst, line 60: can you change this so we only extract `jwasm` (and same for Dockerfile if appropriate)?

PR #1420, opened 2025-01-28
Fix local environment for ARM-based hosts

macabeus:
I had a couple of issues when trying to set up the environment for my ARM computer (MacBook with M1 Max chip).

This PR updates the `Dockerfile` and some other scripts to make it work straightforward for those who use ARM.


backend/Dockerfile
```diff
@@ -33,7 +33,11 @@ RUN git clone "https://github.com/google/nsjail" --recursive --branch 3.1 /nsjai
 
 FROM base AS build
 
-RUN apt-get -y update && apt-get install -y \
+RUN apt-get -y update && \
+    apt-get install -y \
+    libpq-dev \
+    libffi-dev \
+    libssl-dev \
```
macabeus, line 40: These new packages are necessary to build `psycopg2`.
We can't use the binary version since they won't run.

backend/docker_entrypoint.sh
```diff
@@ -8,15 +8,20 @@ BE_PORT=${BACKEND_PORT:-8000}
 
 poetry config virtualenvs.path /backend/virtualenvs
 
+poetry lock --no-update
```
macabeus, line 11: It isn't mandatory, but I think it improves the dx, since whenever we update `backend/pyproject.toml` now it'll install the new packages when restarting the container, without any additional commands.

backend/docker_entrypoint.sh
```diff
@@ -8,15 +8,20 @@ BE_PORT=${BACKEND_PORT:-8000}
...
-  regedit $reg
-done
+if command -v regedit &> /dev/null; then
+  for reg in /backend/wine/*.reg; do
+    echo "Importing registry file $reg..."
+    regedit $reg
+  done
+else
+  echo "regedit command not found. Skipping registry import."
+fi
```
macabeus, line 24: We don't need to run `regedit` if this command isn't available (e.g., when running it from ubuntu).

backend/pyproject.toml
```diff
@@ -28,6 +28,7 @@ django-session-timeout = "^0.1.0"
 django = "^5.1.3"
 requests = "^2.32.3"
 setuptools = "^75.6.0"
+cffi = { version = "^1.17.1", markers = "sys_platform != 'win32'" }
```
macabeus, line 31: The same case as `psycopg2`: we need to build it

mkst:
Is it not enough to set the platform - i.e. just adding `--platform=linux/amd64`? I don't understand why you would need to compile anything (unless you were trying to use arm native ubuntu 22.04).

macabeus:
@mkst Actually, we don't need it 😅
It may be because of a cache issue when I firstly installed the packages without `--platform=linux/amd64`.
It worked well now, after I deleted `virtualenv` and tried to run the precompiled packages.

I removed the unnecessary PR changes, then.

mkst (approved):

PR #1426, opened 2025-02-01
Add warning if user is creating scratch with the same name as an exis…

### LINKED ISSUE #1425: Tell user if a scratch with the same name already exists
If a user creates a scratch, and a 2nd user creates a scratch with the same asm *but* with some symbols renamed, the asm is different and so they won't be part of the same family - meaning user 2 may waste time matching a scratch that has already been matched.

It should be fairly easy to check if any scratches have the same platform+diff label (could limit the check to the Preset if the user has specified one) when a user is on the new scratch page.

mkst:
…ting one...

image

Styling can be tweaked if necessary, but i think this does the job. Only returns 10 (page_size) results, but we can tweak that if it becomes an issue.. this is just meant to improve the UX and prevent two people from working on the same function for the same game without realising :)

Closes #1425.

### DIFF
--- a/frontend/src/app/(navfooter)/new/NewScratchForm.tsx
+++ b/frontend/src/app/(navfooter)/new/NewScratchForm.tsx
@@ -17,8 +17,11 @@ import { scratchUrl } from "@/lib/api/urls";
 import basicSetup from "@/lib/codemirror/basic-setup";
 import { cpp } from "@/lib/codemirror/cpp";
 import getTranslation from "@/lib/i18n/translate";
+import { get } from "@/lib/api/request";
 
 import styles from "./new.module.scss";
+import type TerseScratch from "@/lib/api/types";
+import { SingleLineScratchItem } from "@/components/ScratchList";
 
 function getLabels(asm: string): string[] {
     const lines = asm.split("\n");
@@ -68,6 +71,8 @@ export default function NewScratchForm({
     const [diffFlags, setDiffFlags] = useState<string[]>([]);
     const [libraries, setLibraries] = useState<Library[]>([]);
     const [presetId, setPresetId] = useState<number | undefined>();
+
+    const [duplicates, setDuplicates] = useState([]);
 
     const [ready, setReady] = useState(false);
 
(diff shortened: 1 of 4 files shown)


frontend/src/app/(navfooter)/new/NewScratchForm.tsx
```diff
@@ -257,6 +262,34 @@ export default function NewScratchForm({
...
+                // otherwise filter on platform
+                return scratch.platform === platform;
+            });
+        };
+
+        get(`/scratch?search=${label}`)
+            .then((x) => x.results)
+            .then(filterCandidates)
+            .then(setDuplicates);
+    }, [label, platform, presetId]);
```
bates64, line 291: Since label is a text input, every character the user types into it will trigger a network request, which we don't want. I'd probably wrap use `useDebounce`, like this code that will only make the api call 1000ms after the user finishes typing: https://github.com/decompme/decomp.me/blob/ae0529095372de677e9f6660aab8328dd55c5f95/frontend/src/components/Scratch/panels/DecompilePanel.tsx#L24-L39

frontend/src/app/(navfooter)/new/new.module.scss
```diff
@@ -113,3 +113,14 @@
         }
     }
 }
+
+.duplicatesContainer {
```
bates64, line 117: Probably use tailwind instead given #1394

frontend/src/components/ScratchList.tsx
```diff
@@ -281,9 +284,34 @@ export function ScratchItemPresetList({
     );
 }
 
+export function ScratchOwnerAvatar({ scratch }: { scratch: api.TerseScratch }) {
+    return (
+        scratch.owner &&
+        (!api.isAnonUser(scratch.owner) ? (
+            userAvatarUrl(scratch.owner) && (
```
bates64, line 291: This could be more clearly written without ternaries: `if (scratch.owner && ...) { return <Image /> } else { return <Frog /> }`. Great idea for a component

bates64 (changes requested):

frontend/src/components/ScratchList.tsx
mkst: I just lifted this code verbatim from the Search code, but yeh that makes sense

mkst: Can't appease the linter so gonna leave this as-is.

PR #1435, opened 2025-02-05
Try using SWRImmutable for compilers, platforms, presets etc

mkst:
creating a PR so i can try this out in vercel.. seemed to do as expected locally.

### DIFF
--- a/frontend/src/components/Scratch/SortableFamilyList.tsx
+++ b/frontend/src/components/Scratch/SortableFamilyList.tsx
@@ -3,7 +3,7 @@ import { useMemo, useState } from "react";
 import Link from "next/link";
 
 import classNames from "classnames";
-import useSWRImmutable from "swr/immutable";
+import useSWR from "swr";
 
 import { get } from "@/lib/api/request";
 import type { TerseScratch } from "@/lib/api/types";
@@ -18,12 +18,11 @@ import Sort, {
 import UserLink from "../user/UserLink";
 
 function useFamily(scratch: TerseScratch) {
-    const { data: family } = useSWRImmutable<TerseScratch[]>(
+    const { data: family } = useSWR<TerseScratch[]>(
         `${scratchUrl(scratch)}/family`,
         get,
         {
             suspense: true,
-            refreshInterval: 1000 * 60 * 5, // 5 minutes
         },
     );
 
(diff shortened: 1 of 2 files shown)


ethteck:
Some of these things, particularly presets are actually mutable though. I think I'd feel a bit better if we reduced the update time to a minute rather than an hour for those. For the things not in the db (compilers, platforms) I don't think we need to ever revalidate the data, since the server will reboot anyway when those things change

frontend/src/components/Scratch/SortableFamilyList.tsx
```diff
@@ -18,11 +18,12 @@ import Sort, {
 import UserLink from "../user/UserLink";
 
 function useFamily(scratch: TerseScratch) {
-    const { data: family } = useSWR<TerseScratch[]>(
+    const { data: family } = useSWRImmutable<TerseScratch[]>(
```
ethteck, line 21: I'd think we want families to update instantaneously

frontend/src/lib/api.ts
```diff
@@ -52,7 +53,10 @@ export * from "./api/request";
 export * from "./api/types";
 
 export function useThisUser(): User | AnonymousUser | undefined {
-    const { data: user, error } = useSWR<AnonymousUser | User>("/user", get);
+    const { data: user, error } = useSWRImmutable<AnonymousUser | User>(
```
ethteck, line 56: Usernames can change, though it is rare.. wondering if anything else can change that might be important...

frontend/src/lib/api.ts
```diff
@@ -393,8 +397,8 @@ export function usePresets(platform: string): Preset[] {
     };
 
     const url = typeof platform === "string" ? "/preset" : null;
-    const { data } = useSWR([url, platform], getByPlatform, {
-        refreshInterval: 0,
+    const { data } = useSWRImmutable([url, platform], getByPlatform, {
+        refreshInterval: 1000 * 60 * 60, // 1 hour
```
ethteck, line 401: this one and usePreset I think should be reduced a lot if not reverted to be a normal useSWR thing

frontend/src/lib/api.ts
mkst: User endpoint is `mutate`'d when a user logs out, so this is ok.

PR #1440, opened 2025-02-07
Ensure parents are considered family 👪

mkst:
I think this works...


ethteck:
Thank you for doing this so quickly lol. The issue I see here, though, is that it doesn't get siblings with changed diff_labels. just parents. maybe this is fine enough for now? but it should be noted somewhere.

agh idk, it does feel a bit weird for the family endpoint to be asymmetrical, depending on who you start from

PR #1441, opened 2025-02-08
Simplify backend dockerfile

mkst:
Just install all packages/requirements by default. The downloading of compilers is still managed via the environment vars (i.e. clang is disabled by default), so this just keeps things simple.


mkst:
Created https://github.com/dosemu2/dosemu2/issues/2377 for dosemu2 failures...

mkst:
... And they've already pushed a fix. These guys work fast!

PR #1442, opened 2025-02-08
Don't create profiles for YandexRenderResourcesBot

mkst:


ethteck:
looks grood. I think next time we add something, we should just have a list of things and iterate through the list though

PR #1443, opened 2025-02-08
Make ScratchList loading less greedy

### LINKED ISSUE #1250: When changing the sort method on a scratch list, the 'loading' ui is too greedy
I would expect the "Scratches" heading and selection box to stay during the loading and only the scratch list to switch to the loading 'screen'
image

But instead the loading area makes these ui elements inaccessible
image

Also, if we could have a nicer-looking suspense here, that'd be cool. Something like this, but for a scratch list:
image

mkst:
Closes #1250.


mkst:
Looks like this (apologies for light theme)

image

PR #1453, opened 2025-02-10
Add extra sanity check to user.is_online()

mkst:
So, I believe that `last_request_date` only actually set on profile.save() ... and for the Yandex bot that trawls/executes javascript we create a generic profile but do not save it.


mkst:
Thanks mypy.

PR #1459, opened 2025-02-11
Stick thousands separators in num_scratches on preset page

mkst:
We have some presets with thousands of scratches, so stick a `,` in the number to avoid `3596` like below:

image

### DIFF
--- a/frontend/src/app/(navfooter)/platform/[id]/page.tsx
+++ b/frontend/src/app/(navfooter)/platform/[id]/page.tsx
@@ -65,7 +65,7 @@ export default async function Page({ params }: { params: { id: number } }) {
                     url={`/scratch?platform=${platform.id}&page_size=20`}
                     item={ScratchItemPlatformList}
                     isSortable={true}
-                    title={`Scratches (${platform.num_scratches})`}
+                    title={`Scratches (${platform.num_scratches.toLocaleString("en-US")})`}
                 />
             </section>
         </main>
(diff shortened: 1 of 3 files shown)


frontend/src/app/(navfooter)/platform/[id]/page.tsx
```diff
@@ -65,7 +65,7 @@ export default async function Page({ params }: { params: { id: number } }) {
                     url={`/scratch?platform=${platform.id}&page_size=20`}
                     item={ScratchItemPlatformList}
                     isSortable={true}
-                    title={`Scratches (${platform.num_scratches})`}
+                    title={`Scratches (${platform.num_scratches.toLocaleString("en-US")})`}
```
Comment (bates64) [inline frontend/src/app/(navfooter)/platform/[id]/page.tsx:68]: Forcing en-US here may confuse some Europeans, where 1,000 means our 1.000. I assume you did this due to SSR - either we ignore those warnings or use the request locale. There's probably a Next.js thing for this

Comment (mkst) [reply frontend/src/app/(navfooter)/platform/[id]/page.tsx:68]: I did "en-US" because thats what we have other places:

https://github.com/decompme/decomp.me/blob/de5bd162d31b1899afacbab6c03577604e362f27/frontend/src/app/(navfooter)/platform/%5Bid%5D/page.tsx#L30

https://github.com/decompme/decomp.me/blob/de5bd162d31b1899afacbab6c03577604e362f27/frontend/src/app/(navfooter)/preset/%5Bid%5D/page.tsx#L31

So I was trying to be consistent :)

Comment (ethteck) [reply frontend/src/app/(navfooter)/platform/[id]/page.tsx:68]: I made https://github.com/decompme/decomp.me/issues/1463 to address this for later

PR #1474, opened 2025-02-17
Add initial support for AI ✨

macabeus:
This PR adds the initial support for AI.

It's the first phase, and there are many improvements that can be done later - I wrote some of them at the end.

Although it's a simple integration, it's powerful.
It helps us to write good prompts that can achieve more than 90% match when using reasoning model.
Also, releasing it will be useful to get user's feedback and learn how to improve this feature.

The proposed flow and changes are the following:

1. Now there is a new checkbox when creating a scratch. It's unchecked by default. When checked and there is no AI API Key set, it shows a message asking for the user set one.
image

2. On settings, there is a new page for AI. For now, I added only OpenAI and DeepSeek. Since I'm using [AI SDK](https://sdk.vercel.ai/docs/foundations/providers-and-models), it's easy to add as many providers as necessary, and we could even use a text field for model. But I'm starting with only these settings because it's easy to start with.

image

3. After defining an API Key, the user can go back to the "New scratch" form and see more options. They can copy and paste a scratch link to use as an example in the prompt, and text freely the prompt, using some templates and following the suggestions. 

image

5. There is a new panel in the scratch page: AI. It's displayed only if there is an AI API Key set. This panel is a chat-based integration with the AI. The first message in the chat is the one written in the "New scratch" form.

image

And that's all!

Some improvement that I'm planning to do later:
- improve the code quality for `NewScratchForm.tsx`
- add proper syntax highlighter for the prompt and on the chat panel
- add [hover tooltips](https://codemirror.net/examples/tooltip/) when hovering a template
- add templates for the chat input (e.g., to include the mismatches lines)
- automatically suggest a scratch to use as an example (probably using [[Locality Sensitive Hashing](https://en.wikipedia.org/wiki/Locality-sensitive_hashing)](https://en.wikipedia.org/wiki/Locality-sensitive_hashing) as [Sentry does](https://blog.sentry.io/how-sentry-decreased-issue-noise-with-ai))


ethteck:
Thanks for all your efforts in pulling this together. I can try to get to the code this week, but in the meantime, I have a few suggestions

> Now there is a new checkbox when creating a scratch. It's unchecked by default. When checked and there is no AI API Key set, it shows a message asking for the user set one.
If there isn't an API key set, the box should remain disabled. It shouldn't be clickable at all without a key, imo. The message you have currently about needing to add the key can be hover-text / also atted to the description of the text box after the existing sentence

Regarding the text about how reasoning models are more expensive, I think we should simply say that we recommend reasoning models because they are best suited for the task of decompilation, but to be mindful of pricing (and maybe link to each provider's pricing page somewhere).

red031000:
how does this actually work with decomp matching?
I have in the past tried to use AI for decomp (chatgpt 3.5) and it has never got even remotely close to decomp, in my experience it treats it like -O0 code, being extremely verbose about what is being done, and never gets anywhere near matching
part of me wonders if this will genuinely be more of a hinderance than a help for decomp, as it can give you the wrong ideas about some things make it very difficult to match
there are also a lot of ethical/moral issues to consider, including electricity use, environment, data the AIs are trained on and rights to them, as well as privacy concerns with some AI models, and lack of control of the data sent etc

macabeus:
@ethteck Sure, I'll apply your suggestions. Thanks!

<hr />

@red031000 

That's the magic of reasoning model!
Until a few months ago, there was no available reasoning model available for us, but now we can use freely reasoning models such as OpenAI's `o1-preview` and DeepSeek's R1, and they work wondrously well - even with a simple prompt and no fine-tuning.

Of course, we still need to manually improve the code quality even if you get automatically a 100% match - but it's a great quick start.

[You can check this PR](https://github.com/SAT-R/sa3/pull/4) to have a quick glance at how it works and the result. It was my first PR working in a decomp project and it helped a lot.
It wasn't a hinderance. Instead, it helped a lot, engaged me, and surely, it can be useful for others too.
I also shared a video a few days ago demoing this feature. [You can check see it here](https://discord.com/channels/897066363951128586/897075447534354462/1338697900649353289).

Furthermore, it's important to highlight this PR introduces a strictly optional feature which depends on users providing their own API keys, on the provider that they prefer. It allows each person to make an informed decision on whether they wish to use AI themselves or not.
In a future update, I would also like to provide support for local AI models (see [`#website-help`](https://discord.com/channels/897066363951128586/897077740476448768/1340407468718559384)) so that users can further avoid privacy concerns.
The goal is to empower users with flexible options.

frontend/src/app/(navfooter)/settings/ai/AISettings.tsx
```diff
@@ -0,0 +1,91 @@
...
+};
+const deepSeekRecommendedModel = AIModel.DEEPSEEK_REASONER;
+
+export default function AISettings() {
+    const [aiProvider, setAIProvider] = useAIProvider();
+    const [aiModel, setAIModel] = useAIModel();
+    const [aiAPIKey, setAIAPIKey] = useAIAPIKey();
+
+    const modelOptions =
+        aiProvider === AIProvider.OPEN_AI ? openAIModels : deepSeekAIModels;
```
bates64, line 49: Better to use an object to look this up, since we are likely to add providers in future

frontend/src/app/(navfooter)/settings/ai/AISettings.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+    return (
+        <>
+            <Section title="Provider">
+                <RadioList
+                    value={aiProvider}
+                    onChange={(value) => {
+                        setAIProvider(value as AIProvider);
+                        setAIModel(
+                            value === AIProvider.OPEN_AI
```
bates64, line 59: Ditto here. You may want to make the object map provider to { models, recommendedModel }

bates64 (changes requested):
Cool!!

frontend/src/app/(navfooter)/new/NewScratchForm.tsx
```diff
@@ -417,6 +481,87 @@ export default function NewScratchForm({
...
+                                Final Prompt
+                            </FormLabel>
+
+                            {/* TODO: Make it be readonly */}
+                            <CodeMirror
+                                className="w-full flex-1 overflow-hidden rounded border border-[color:var(--g500)] bg-[color:var(--g200)] [&_.cm-editor]:h-full"
+                                value={finalPrompt}
+                                valueVersion={finalPrompt as any}
+                                extensions={[basicSetup]}
+                            />
```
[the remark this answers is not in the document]
bates64: If you want syntax highlighting for the prompt markdown, then CodeMirror is right-- but make sure readonly is added before this is merged, at least

bates64:
A future iteration of this could be to perform neural decompilation _on existing scratches_. So, the agent would be given access to all the tools the user gets and can try to beat the current match%, a bit like the permuter.

I also recommend looking into
- Neural symbol naming (there's a 2023/24 paper on this but I can't seem to find it now)
- https://www.ndss-symposium.org/wp-content/uploads/2024-401-paper.pdf which introduces a three-role mechanism
- https://arxiv.org/abs/2501.04811 (haven't read, but is recent and sounds like an interesting new match% algorithm appropriate for AI)

ethteck:
Upon thinking about this further and discussing with @mkst , I think I'd really prefer if AI was not part of the scratch creation page. I think it makes the most sense to think about as a decompiler, akin to how m2c is one of our supported decompilers.

mkst:
For the platforms where we don't have a decompiler (heck, even for the ones we do), then passing asm + some "context" + a well engineered prompt to an AI is certainly an interesting idea. How well do these models do on raw asm alone? How do they compare to non-AI, off-the-shelf decompilers like Ghidra or IDA?

In its current form I think this is quite niche (user will need to have an API token) and appears quite slow. I feel like it could be a separate tool - *outside* of decomp.me - and perhaps once mature (i.e. a) no API keys required, b) responses come back in a few seconds) we look to integrate it with the site.

I also don't know how fragile it is - my concern is that if something breaks there may not be sufficient interest/usage to fix it (we've had this in the past with platforms being added, not used, breaking, and ultimately being removed as there was no-one motivated to fix).

bates64:
> responses come back in a few seconds

I doubt this is feasible in the medium term. The reasoning models need a lot longer to think, and our use of it is multi-shot so if it's a difficult function to crack it might take even longer. I don't see that changing anytime soon.

It's probably feasible to make this into a library that we call into, just like mips2c.

macabeus:
@bates64 I like these next steps. I didn't know the term "neural decompilation" and these papers.

We can talk later about it, but _I think_ that the best quick win to do next for AI prompting is automatically selecting similar scratches to include as examples. We can use [Locality-sensitive hashing](https://en.wikipedia.org/wiki/Locality-sensitive_hashing) (LSH) to group the scratches when creating a new one, [as Sentry does to group the issues](https://blog.sentry.io/how-sentry-decreased-issue-noise-with-ai/).

It's nice because it's useful even for non-AI users, since they can learn how others similar functions were decompiled, and recognize the patterns. Adding a dropdown on the "About" or "Family" tabs to see similar scratches seems good.

<hr />

@ethteck I didn't know about the "Decompile" panel. Nice! I worked only for GBA, and it doesn't have this feature yet.

But note that this PR didn't add a new _decompiler_, rather, it adds a _chat_.
Although the current flow encourages the first message to works like a decompiler, it's only the first phrase.
I'm planning to improve the chat to include more iterations (e.g., allow asking AI to fix the mismatch in the selected line, something as GitHub Copilot Chat does on VS Code).
I did some tests in DeepSeek Chat, and LLM works well not only to generate the first decompiled code, but also to _increase_ the matching.

Moving it to a new tab has the advantage of having more space for the chat, and it would fit well the form currently from "New scratch".
On the other hand, the downside is a context switch (I can't see the chat and the compilation at the same time).

I need to think more about how it would imply the UX and next steps, the tradeoff in both approaches... 🤔

<hr />

@mkst LLM works much better!
In my experience, when working on Sonic Adventure 3, Ghidra generates an almost useless C code.
When I tested using Claude, still without a reasoning model, it returned a better result than Ghidra, but it's still only _good_.
And finally, reasoning models are a game changer. It produces a _great_ result.
I also did some tests using only the asm in the prompt, and it works well too. Of course, with a better prompt, it gets much better results. My changes on `NewScratchForm.tsx` encourage the users to write a good prompt.

I agree that this PR adds a quite niche feature - but it's the first phrase.
As @bates64 wrote above, I don't think we'll be able to run it fast or without an API key in the short/medium terms. Reasoning models requires too much memory and processing that I don't think it'll become feasible to run on the client-side or on decomp.me's backend (well, AI is evolving quickly, so I may be wrong =P).
In any case, even if it currently requires an AI key and takes 1–2 minutes to respond, I think it's good enough to see a few users using it and collecting feedback to learn how should be the next steps (better chat templates? adding a provider that produces faster result? better setup instructions? etc).

Also, I'm not sure how to turn it into a library, since I updated a couple of files to can integrate it into the site (e.g., settings, UI for the scratch page).
But I can surely improve the code quality to make it more isolated (my first next step is improving the code quality for `NewScratchForm.tsx`, since it's a messy now 😅).

frontend/src/lib/ai/prompt.ts
```diff
@@ -0,0 +1,77 @@
...
+Compiled to the last function of this code:
+\`\`\`
+${exampleSource}
+\`\`\``
+            : "",
+        decompileRequestSection: `${hasExample ? "Given the above context, translate" : "Translate"} this assembly to an equivalent ${targetLanguage} code:
+\`\`\`
+${sourceAsm}
+\`\`\``,
+        compilerNotes: compilerNotes[compilerId] ?? "",
```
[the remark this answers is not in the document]
macabeus: For now, I think that only the compiler id is enough to get its related notes.

backend/coreapp/serializers.py
```diff
@@ -209,6 +209,9 @@ class ScratchSerializer(serializers.ModelSerializer[Scratch]):
     preset = serializers.PrimaryKeyRelatedField(
         required=False, allow_null=True, queryset=Preset.objects.all()
     )
+    target_assembly_source_asm = serializers.CharField(
```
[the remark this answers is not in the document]
bates64: I misread this as the TerseScratchSerializer. Should be fine as-is

bates64:
o1-preview has been released as o1, and o3-mini has recently been released and is 1/15th of the price. We should test it out and, if it works well, recommend it for OpenAI users.

frontend/src/components/Chat/Chat.tsx
```diff
@@ -0,0 +1,65 @@
...
+        );
+    }
+
+    return (
+        <div className="h-full">
+            {chatHistory.length === 0 && <ChatEmptyState />}
+
+            <div ref={chatHistoryRef} className="h-auto">
+                {chatHistory.map((message, i) => (
+                    <div key={i}>
```
bates64, line 51: Don't use `i` as a key; it's not stable between renders. https://react.dev/learn/rendering-lists#where-to-get-your-key

frontend/src/components/PromptBuilder/PromptBuilder.tsx
```diff
@@ -0,0 +1,169 @@
...
+                    extensions={[basicSetup]}
+                />
+
+                <div className="mt-2 text-[0.8rem] text-[color:var(--g800)]">
+                    <p>Suggestions:</p>
+
+                    {promptSuggestions.size === 0 && <p>- None</p>}
+
+                    {Array.from(promptSuggestions).map((suggestion, i) => (
+                        <p key={i} className="pl-2">
```
bates64, line 117: Ditto. Probably just use `suggestion` as the key here

frontend/src/components/PromptBuilder/PromptBuilder.tsx
```diff
@@ -0,0 +1,169 @@
...
+                    onChange={(value) => {
+                        setRawPrompt(value);
+                    }}
+                    extensions={[basicSetup]}
+                />
+
+                <div className="mt-2 text-[0.8rem] text-[color:var(--g800)]">
+                    <p>Suggestions:</p>
+
+                    {promptSuggestions.size === 0 && <p>- None</p>}
```
bates64, line 114: Probably just hide the entire suggestions div

macabeus:
@bates64 
> o1-preview has been released as o1, and o3-mini has recently been released and is 1/15th of the price. We should test it out and, if it works well, recommend it for OpenAI users.

I really would like to test them, but they aren't available for everyone yet 😞
`o1` and `o3-mini` are available only for a subset of tier 3+ users
https://help.openai.com/en/articles/10362446-api-access-to-o1-and-o3-mini

frontend/src/components/Chat/Chat.tsx
macabeus: I think it's one of the cases that it's okay to use the array index: the array is never reordered, no elements are deleted, and new elements are always added at the end.
But okay, I added a message id.

PR #1480, opened 2025-03-07
Experimental objdiff-web integration

encounter:
This replaces the existing objdiff experiment with one based on [objdiff-web](https://github.com/encounter/objdiff-web). When enabled, "objdiff [alpha]" shows up as a new tab alongside the regular Compilation tab.

I'll still have to figure out how to best integrate it with the decomp.me build/deployment. In the meantime, `OBJDIFF_BASE` points to a hosted version at https://diff.decomp.dev/. This is not intended to be permanent, but was the path of least resistance for testing.

For development purposes, one can setup the objdiff-web repo and uncomment the line in .env that points it to localhost. Live reload and everything works seamlessly.

Screenshot From 2025-03-06 23-10-19
Screenshot From 2025-03-06 23-10-14
Screenshot From 2025-03-06 23-10-01

It inherits the color scheme, font and font size from the decomp.me settings:
image


frontend/src/components/ScoreBadge.tsx
```diff
@@ -8,7 +8,7 @@ export function calculateScorePercent(score: number, maxScore: number): number {
         return 0;
     }
 
-    if (maxScore === 0) {
+    if (!maxScore) {
```
ethteck, line 11: Is this a biome thing? Checking the maxScore is zero seems much more readable than checking if not maxScore

encounter: It’s intended to handle null/undefined instead of just 0. This was fixing some runtime errors I hit while messing with stuff, iirc

encounter: But I should clarify it’s not necessary for this set of changes, it just snuck in from local testing I was doing. Happy to revert if desired

ethteck: ohh okay. that's fine, I didn't realize the value was ever not numeral

PR #1482, opened 2025-03-10
Update privacy policy

ethteck:
Opening this as a PR for a bit and will merge if no one has any issues. This isn't much of a change in policy other than to remove the notion we won't share data with "third parties" because it's ambiguous what a third party would be, and it's also untrue that people not associated with the site should be able to access site data. 

For example, researchers wanting to improve decompilers by looking at the scratch database could easily be considered a third party, but this is also a totally valid use case for handling site data.

To reiterate, all  the data we're talking about is already publicly available on the site, visually and through the API. Nothing about the data in question involves private user info, authentication tokens, or anything of that nature.


bates64 (approved):
LGTM, good idea. We may need to issue an email to all users of the privacy policy update, I'm unsure what the laws are like here

PR #1485, opened 2025-03-12
Update asm-differ (x86 reloc fix)

1superchip:
https://github.com/simonlindholm/asm-differ/pull/186


mkst:
Replaced by https://github.com/decompme/decomp.me/pull/1486.

PR #1489, opened 2025-03-16
Add 4 new versions of SN 2.7.2 for Nintendo 64

galaxyhaxz:
This adds 2 previously lost SN 2.7.2 versions and their C++ versions, totalling 4 new compilers for the N64. The previous version of SN 2.7.2 has also been renamed to reflect its build compared to these builds. Note that for build 0001 to work, a recently updated `psyq-obj-parser` is needed to handle the GP relocations. The cygnus builds (0004/6) disable `-G` by default and even seem to block its usage, so `-G0` seems fine to use for them.

Also note that the compilers currently have no download links, this should be updated once they have a home.

### DIFF
--- a/backend/coreapp/compilers.py
+++ b/backend/coreapp/compilers.py
@@ -882,8 +882,8 @@ CCN64_CPP_CXX = '/usr/bin/cpp -E -lang-c++ -undef -D__GNUC__=2 -D__cplusplus -Dm
 GCC272SN0001 = GCCCompiler(
     id="gcc2.7.2sn0001",
     platform=N64,
-    cc=CCN64_CPP_C +
-    '| ${WIBO} "${COMPILER_DIR}"/cc1n64.exe ${COMPILER_FLAGS} -o "$OUTPUT".s '
+    cc=CCN64_CPP_C
+    + '| ${WIBO} "${COMPILER_DIR}"/cc1n64.exe ${COMPILER_FLAGS} -o "$OUTPUT".s '
     '&& ${WIBO} "${COMPILER_DIR}"/asn64.exe -q "$OUTPUT".s -o "$OUTPUT".obj '
     '&& "${COMPILER_DIR}"/psyq-obj-parser "$OUTPUT".obj -o "$OUTPUT" -b -n -s',
 )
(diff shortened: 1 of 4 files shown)


mkst:
Are you good to make a PR for the [compilers](https://github.com/decompme/compilers) repo? if you attach all the zips I can attach them to the Releases page (like many other compilers)?

galaxyhaxz:
Yep, I'm working on that now!

mkst:
If you're renaming `gcc2.7.2sn` to `gcc2.7.2sn0001` you'll also need a migration script to update all the scratches in the database. 

An example would be https://github.com/decompme/decomp.me/blob/main/backend/coreapp/migrations/0057_rename_ps1_mipsel_compilers.py

galaxyhaxz:
Done. I don't think there are any presets which used it but that's added just in case. Also I'm not sure if the C++ versions need to be listed in `compilers.linux.yaml` or not, at least the 2.8.1 SN C++ isn't there.

backend/coreapp/compilers.py
```diff
@@ -876,12 +876,50 @@ def available_platforms() -> List[Platform]:
     cc='"${COMPILER_DIR}"/gcc -G0 -c -B "${COMPILER_DIR}"/ $COMPILER_FLAGS "$INPUT" -o "$OUTPUT"',
 )
 
-GCC272SN = GCCCompiler(
-    id="gcc2.7.2sn",
+GCC272SN0001 = GCCCompiler(
+    id="gcc2.7.2sn0001",
+    platform=N64,
+    cc='/usr/bin/cpp -E -lang-c -undef -D__GNUC__=2 -Dmips -D__mips__ -D__mips -Dn64 -D__n64__ -D__n64 -D_PSYQ -D__EXTENSIONS__ -D_MIPSEB -D__CHAR_UNSIGNED__ "$INPUT" '
```
mkst, line 882: are the `cc` commands the same across all versions? can this be pulled out into a variable to avoid the duplication?

galaxyhaxz: The cpp flags are the same, yeah, I just copied them from 2.8.1 SN. The other commands are slightly different--0001 uses the default G flag (-G8) whereas 0006 specifies 0001. For dealing with C++, I don't think you can pass a flag to CCN64 to force it to be C++ which is why it has to manually call `cpp` and pass it to the C++ compiler instead of the C one.

mkst:
I think you just needed a `+` to concatenate the variable with a string, whereas there's implicit string-string concatenation

galaxyhaxz:
You can tell I'm a python noob :P Hopefully now the tests will pass.

PR #1495, opened 2025-03-19
Update default font & increase size

encounter:
This pulls in [JetBrains Mono](https://www.jetbrains.com/lp/mono/) as a variable web font (2 KiB) and uses it as the default font family. The default size is updated from 11px to 13px to improve legibility.

Additionally, an appearance setting "Enable ligatures" is added, allowing users to disable the font ligatures that are included in JetBrains Mono (or their own configured font).

<p align="center">Before / After (Appearance Settings)</p>
<p align="center">
  Before
  After
</p>

<p align="center">Before / After (Scratch)</p>
<p align="center">
  Before
  After
</p>

Note that my default `monospace` font is configured to `Noto Sans Mono`, which is similar to JetBrains Mono, so you won't see much difference in the font in these comparisons. It will be more apparent if the monospace font is not configured at a system level.

Rationale:
- Currently, the default font-family is `Menlo, Monaco, monospace`, which are macOS fonts. This means most users will fall back to `monospace`, which is up to the user agent to choose. Instead, we can unify this across platforms by pulling in a small (2 KiB) web font.
- 11px is very small and hurts readability. 13px is a more reasonable default.

FYI, I'm not committed to JetBrains Mono in particular, if anyone has other suggestions.

### DIFF
--- a/frontend/src/app/(navfooter)/credits/page.tsx
+++ b/frontend/src/app/(navfooter)/credits/page.tsx
@@ -14,12 +14,14 @@ const OTHER_PROJECTS = {
     m2c: "https://github.com/matt-kempster/m2c",
     "psyq-obj-parser":
         "https://github.com/grumpycoders/pcsx-redux/tree/main/tools/psyq-obj-parser",
+    objdiff: "https://github.com/encounter/objdiff-web",
     Django: "https://www.djangoproject.com/",
     "Django REST Framework": "https://www.django-rest-framework.org/",
     "Next.js": "https://nextjs.org/",
     React: "https://reactjs.org/",
     "Tailwind CSS": "https://tailwindcss.com/",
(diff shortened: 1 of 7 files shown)


ethteck:
looks great! I'm neither here nor there on the size change (I don't use the size zoomed in), but if it helps others and 11 is considered small, happy to change the default

ethteck:
oh, two other things:

* it'd be nice of us to put jetbrains mono in the credits page (though not required)
* I'd prefer liguatures off by default. there's been plenty of times in decomp projects where I have to pay very specific attention to the characters in a string, and having them be represented in a way that's easier to read but maybe makes them look like one character could be distracting in those cases

encounter:
Sounds good, I disabled ligatures by default & updated the credits page. (I added objdiff-web there too while I was at it)

PR #1500, opened 2025-03-21
Replace unmaintained `use-persisted-state`

encounter:
Follow up to #1499. For some reason, I decided to write a React hook to replace `use-persisted-state`. I think it's pretty neat. I'll definitely be copying it into my own projects.

https://github.com/user-attachments/assets/cd33fb9a-2f75-4d18-8dd3-731386f279bf

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -53,7 +53,6 @@
         "sass": "^1.82.0",
         "swr": "^2.2.5",
         "use-debounce": "^10.0.4",
-        "use-persisted-state": "^0.3.3",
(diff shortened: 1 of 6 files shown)


mkst:
Hard to see on mobile, did you remove the try/catch that I added around the theme being light/dark/auto -because of the bug in this library? 🤞

mkst:
Realised I might be hallucinating, might have only been something I did locally? 😅

encounter:
I didn’t see a try/catch, if you see any places that worked around a bug in this library let me know!

Issue #1538, opened 2025-05-16
.marker position causes unnecessary horizontal scrollbar in Scratch editor


mkst:
In CodeMirror.tsx we determine the width of the editor and assign it to a css variable

```js
    const { ref: el, width } = useSize<HTMLDivElement>();
    ...
        <div
            ref={el}
            onMouseMove={debouncedOnMouseMove}
            className={clsx(styles.container, className)}
            style={
                {
                    "--cm-font-size": `${fontSize}px`,
                    "--cm-container-width": `${width}px`,
                } as CSSProperties
            }
        />
```

which we use in the CSS for `useCompareExtension.ts`:
```css
.marker {
    &::after {
        content: "";

        position: absolute;
        left: 0;
        right: 0;
        width: var(--cm-container-width); // Set by <CodeMirror> component
        height: calc(var(--code-line-height) * 1em);

        pointer-events: none;

        background: #00ffff10;
    }
}
```

However, when there is a vertical scrollbar in the scratch editor, we end up getting a horizontal one too - because the `.marker` is is being placed 'too far' to the right.

My CSS is poor and ChatGPT couldn't give me a solution that wasnt equivalent to boiling the ocean.. so I'm creating the issue in case someone with a brain wants to fix it!

PR #1539, opened 2025-05-16
Fast return from useCompareExtension when compareTo is undefined

mkst:
This is a slight band-aid for #1538. The issue of unnecessary horizontal scrollbars won't occur when there is no parent scratch to compare the current scratch / context to.

### DIFF
--- a/frontend/src/lib/codemirror/useCompareExtension.ts
+++ b/frontend/src/lib/codemirror/useCompareExtension.ts
@@ -139,6 +139,8 @@ export default function useCompareExtension(
 ): Extension {
     const [compartment] = useState(new Compartment());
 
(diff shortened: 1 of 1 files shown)


frontend/src/lib/codemirror/useCompareExtension.ts
```diff
@@ -139,6 +139,8 @@ export default function useCompareExtension(
 ): Extension {
     const [compartment] = useState(new Compartment());
 
+    if (typeof compareTo === "undefined") return [];
```
bates64, line 142: Put this after all the 'use' calls, hooks shouldn't be conditional

bates64 (changes requested):

Issue #1542, opened 2025-05-17
More efficient use of /api/compilers


mkst:
Currently /api/compilers pulls in all compilers (+flags+presets) across all platforms which is quite a hefty query.

We ought to be able to be more efficient, pull down only compilers for the relevant platform (for example) and/or only get presets or flags when we want them.

We can keep this endpoint around, but create leaner versions to use in the frontend.

mkst:
Closing this as we're smarter now 🧠

PR #1544, opened 2025-05-18
Optimise API calls for platform/compiler/presets

mkst:
Aims to close out #1542.

TL/DR: No longer pull down the huge `/compiler` blob, be smarter and pull down the compilers we need for the selected platform. The `/compiler` endpoint still exists, unchanged, for anyone using the API directly, but we don't use it...

Changes:

### `/platform` returns simple platform list:
```json
{
    "dreamcast": {
        "id": "dreamcast",
        "name": "Dreamcast",
        "description": "SH4 (little-endian)",
        "arch": "sh4",
        "has_decompiler": false
    },
    "gba": {
        "id": "gba",
        "name": "Game Boy Advance",
        "description": "ARMv4T",
        "arch": "arm32",
        "has_decompiler": false
    },
    /* etc */
```
### NEW `api/compiler/${platform}` returns available compilers for a specific platform
```json
{
    "ido5.3": {
        "platform": "n64",
        "flags": [
            {
                "type": "flagset",
                "id": "ido_opt_level",
                "flags": [
                    "-O0",
                    "-O1",
                    "-O2",
                    "-O3"
                ]
            },
            {
                "type": "flagset",
                "id": "ido_debug_level",
                "flags": [
                    "-g0",
                    "-g1",
                    "-g2",
                    "-g3"
                ]
            },
            /* etc */
```
### NEW  `api/compiler/${platform}/${compiler}` returns compiler specific details
```json
{
    "ido5.3": {
        "platform": "n64",
        "flags": [
            {
                "type": "flagset",
                "id": "ido_opt_level",
                "flags": [
                    "-O0",
                    "-O1",
                    "-O2",
                    "-O3"
                ]
            },
            {
                "type": "flagset",
                "id": "ido_debug_level",
                "flags": [
                    "-g0",
                    "-g1",
                    "-g2",
                    "-g3"
                ]
            },
```
### NEW `api/preset/${id}/name` to return the preset name (lighter than all info including num_scratches when we only need the name)
```json
{
    "id": 109,
    "name": "LEGO Island"
}
```


backend/coreapp/views/preset.py
```diff
@@ -89,6 +88,3 @@ def name_only(self, request: Request, pk: str) -> Response:
         serializer = self.get_serializer(preset)
         return Response(serializer.data)
```
ethteck, line 90: black wants you to remove a line here i think

backend/coreapp/serializers.py
```diff
@@ -63,6 +63,30 @@ class LibrarySerializer(serializers.Serializer[Library]):
     version = serializers.CharField()
 
 
+class PresetNameSerializer(serializers.ModelSerializer[Preset]):
```
ethteck, line 66: for the sake of consistency, I think it'd be good to have both of these be X PresetSerializer. this one could be TinyPresetSerializer or NamePresetSerializer or something like that

backend/coreapp/views/compiler.py
```diff
@@ -14,10 +16,42 @@
...
 
 
-def endpoint_updated(request: Request) -> datetime:
+def endpoint_updated(request: Request, **_: typing.Any) -> datetime:
     return max(Preset.most_recent_updated(request), boot_time)
 
 
+class SingleCompilerDetail(APIView):
+    @condition(last_modified_func=endpoint_updated)
+    def get(
```
ethteck, line 25: Since this function is just returning a list of compilers, I think the last_modified time would be the application startup time. I we have other example funcs that use that?

ethteck (approved):

PR #1548, opened 2025-05-21
Add 'family' slug to track ultimate ancestor for a Scratch

mkst:
The idea here is to avoid the N+1 query when trying to find related scratches. We keep track of the ultimate parent of each scratch upon creation/forking so that it's a simple lookup to find the family.

If the ultimate parent scratch gets deleted, then we'll fall back to using the immediate parent of the scratch. I think this is a fair trade-off in terms of usability vs performance; ideally people don't delete family scratches if they're interested in tracking the family.

### DIFF
--- a/backend/coreapp/migrations/0060_populate_scratch_family.py
+++ b/backend/coreapp/migrations/0060_populate_scratch_family.py
@@ -1,0 +1,65 @@
+import logging
+
+from typing import Any
+
+from django.apps.registry import Apps
+
+from django.db import migrations
+from django.db.backends.base.schema import BaseDatabaseSchemaEditor
+
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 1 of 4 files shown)


backend/coreapp/migrations/0060_populate_scratch_family.py
```diff
@@ -0,0 +1,65 @@
...
+                current = cache[current.slug]
+                break
+        root = current
+        for s in visited:
+            cache[s.slug] = root
+        return root
+
+    updates = []
+
+    scratches = Scratch.objects.select_related("parent").all()
```
ethteck, line 38: I worry this will fetch all scratches and fail on the site db

mkst: I will ask my pal chatgpt!

Issue #1570, opened 2025-05-29
Create Compiler Preset for GTA SA


ifarbod:
I would like to create/update a compiler preset on decomp.me.

- **Preset Name:** `Grand Theft Auto: San Andreas (v1.0 - MSVC 2002)`
- **Preset ID (when updating):** -
- **Platform:** `win32`
- **Compiler ID:** `msvc7.0`
- **Compiler Flags:** `/TP /O2 /G5 /Gs /GS- /MT`
- **Libraries (if applicable):**  `directx9.0`

Example scratch: https://decomp.me/scratch/rjTNv

ifarbod:
Alright, if #1572 is approved, I'd like to update the preset to include dx9!

mkst:
Library now added! Thank you for your PR :)
```json
{
    "id": 175,
    "name": "Grand Theft Auto: San Andreas (v1.0 - MSVC 2002)",
    "platform": "win32",
    "compiler": "msvc7.0",
    "assembler_flags": "",
    "compiler_flags": "/TP /O2 /G5 /Gs /GS- /MT",
    "diff_flags": [],
    "decompiler_flags": "",
    "libraries": [
        {
            "name": "directx",
            "version": "9.0"
        }
    ],
    "num_scratches": 1,
    "owner": null
}
```

ifarbod:
@mkst I'd like to update the compiler flags, didn't want to create a new issue so this is probably the best place for it:

/TP /O2 /Ob2 /Gs /MT

PR #1572, opened 2025-05-30
Add DXSDK 9.0 to win32 libraries

ifarbod:
For #1570 - we need DX 9 in order to compile the code on the website.

### DIFF
--- a/backend/libraries/libraries.yaml
+++ b/backend/libraries/libraries.yaml
@@ -8,3 +8,7 @@ win32:
       git:
         url: https://github.com/roblabla/directx-headers
         branch: main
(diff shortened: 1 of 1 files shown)


mkst:
Is it possible to rename the `Include` directory to `include` or will that mess up includes for you? No worries if it's a problem, I'll just tweak [this line](https://github.com/decompme/decomp.me/blob/main/backend/libraries/download.py#L68) so it's not hardcoded.

mkst:
Cool, yeh I think we just need the `include` directory to be called exactly that. The folders inside can remain as they are 👍

PR #1580, opened 2025-06-05
Add Docker Production setup

mkst:
Creating PR for comments (and so I can give it a review)...


backend/docker_prod_entrypoint.sh
```diff
@@ -0,0 +1,33 @@
...
+DB_HOST=${DATABASE_HOST:-postgres}
+DB_PORT=${DATABASE_PORT:-5432}
+
+BE_HOST=${BACKEND_HOST:-0.0.0.0}
+BE_PORT=${BACKEND_PORT:-8000}
+
+WORKERS=${BACKEND_WORKERS:-4}
+
+# TODO: pull these out of the entrypoint?
+poetry run /backend/compilers/download.py
```
ethteck, line 12: yeah I think these can be manual processes we run at any point. We may want to even decouple them from the current update.sh and just have a dedicated script for them

PR #1611, opened 2025-06-14
Remove .next/cache after build

mkst:


mkst:
How so? does `yarn start` pull down fresh cache?

bates64:
That's beautiful. This is a different panel than I've had. Everything's Docker. That's beautiful. Wow.
Disregard what I said: somehow didn't realise it was a Dockerfile :shipit:

PR #1617, opened 2025-06-15
Warn user on empty diff when using --disassemble=SYMBOL

mkst:
Creating this for comments. This seemed the least-worse place to put the change. Not particularly happy about it... Open to suggestions of how to do this better.


mkst:
Yeh. It works. Just didn't like having to move half of the functions from "run_diff". But I didn't see another way to do it

PR #1623, opened 2025-06-18
Remove 'git' requirement from library downloader

### LINKED ISSUE #1612: [CHORE] Update library download.py to remove dependency on git
Per the title.

We ought to be able to just download the zip of each library repo which would mean we don't need to have `git` in the production backend image which will shave ~70MB off the size.

mkst:
Closes #1612.

### DIFF
--- a/backend/libraries/download.py
+++ b/backend/libraries/download.py
@@ -6,7 +6,10 @@ import functools
 import logging
 import os
 import shutil
-import subprocess
+import requests
+import tempfile
+import zipfile
+import io
 
 from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


backend/libraries/download.py
```diff
@@ -56,9 +133,11 @@ def get_library(
     library_dir.mkdir(parents=True, exist_ok=True)
 
     if "git" in download_info:
-        download_git(
+        res = download_git(
             library_name, library_version, library_dir, download_info["git"], force
         )
+        if res is False:
+            return False
```
ethteck, line 140: isn't this going to happen anyway when we return res at the end? or is there some nuance between none and false I'm missing

mkst: Nuance is meant to be `None` means it wasn't updated, whereas False means it failed

ethteck: ahh I misread the indentation level, nm!

PR #1640, opened 2025-08-09
m2c for ARM

simonlindholm:
... and also use "mipsee" rather than "mipsel" where relevant while we're at it, to get the ABI right.


simonlindholm:
CI fails with:
```
12.50 The following packages have unmet dependencies:
12.60  dosemu2 : Depends: comcom64 but it is not installable
```

PR #1656, opened 2025-09-09
Add optional red/green/blue background to diff rows

### LINKED ISSUE #1633: [IDEA] Set background colour on differing lines
Just opened a scratch and was struggling to find the differences. Thought it'd be nice if "close matches" would switch to a background-colour based diff, like GitHub does. (probably also requires a slightly darker text colour for contrast).

Image

Image

mkst:
Dark
image

Light
image

Closes #1633.


mkst:
Had a look at asm-differ, created [this PR](https://github.com/simonlindholm/asm-differ/pull/199) which 'solves' the problem, but feels hacky, so awaiting some guidance on how to do it in a nicer way!

PR #1666, opened 2025-09-17
add "experimental" SN 2.8.1 C++ compiler (in order to preserve line numbers)

marijnvdwerf:
Made this change in github UI, haven't tested it. Is there a quick way to test if the backend works?

### DIFF
--- a/backend/coreapp/compilers.py
+++ b/backend/coreapp/compilers.py
@@ -897,6 +897,7 @@ GCC281SNCXX = GCCCompiler(
 
 GCC281SNEWCXX = GCCCompiler(
     id="gcc2.8.1snew-cxx",
+    base_compiler=GCC281SN,
     platform=N64,
     cc=CCN64_CPP_CXX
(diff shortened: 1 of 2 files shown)


ethteck:
unfortunately not without running it locally

ethteck:
Could you add a string for this compiler as well? example in this pr: https://github.com/decompme/decomp.me/pull/1658/files

backend/coreapp/compilers.py
```diff
@@ -895,6 +895,14 @@ def available_platforms() -> List[Platform]:
     '&& "${COMPILER_DIR}"/psyq-obj-parser "$OUTPUT".obj -o "$OUTPUT" -b -n',
 )
 
+GCC281SNEWCXX = GCCCompiler(
```
mkst, line 898: you're missing the `base_compiler` key

mkst (commented):
also missing the translation in https://github.com/decompme/decomp.me/blob/main/frontend/src/lib/i18n/locales/en/compilers.json#L74

PR #1671, opened 2025-09-21
cromper

ethteck:
(this is currently unfinished and is in draft state)

This PR adds a new entirely separate service for handling the various processes involves in running the site: assembling, compiling, diffing, and running decompilers.

The backend still talks to this service directly, as well as the frontend server, but now the client can also talk directly to this server to do decompilation tasks, incremental compilations, etc.

There's several reasons we are separating these into a separate service:
* The django server can be freed of many of its responsibilities
* The new service is async (woo), so requests will no longer block each other
* The backend is now much leaner, allowing for possible future rewrites and optimizations without worrying about all this compiler junk

Todo:
- [x] Make compiler, library, and platform endpoints singular
- [x] Deserialize back into proper objects in Django
- [x] Django tests pass
- [x] Fixup frontend `/platform` endpoint to fetch platform data from cromper (or update api.ts to be able to call cromper?)
- [ ] Sort out compiler flags
- [ ] Update nginx config
- [ ] CI passes
- [ ] health checks from django <> cromper?
- [ ] 'super' endpoint to do initial scratch creation in 1 go (TBD: profile first to determine if necessary)


backend/coreapp/cromper_client.py
```diff
@@ -0,0 +1,217 @@
...
+        self,
+        platform_id: str,
+        target_elf: bytes,
+        compiled_elf: bytes,
+        diff_label: str = "",
+        diff_flags: list[str] = [],
+    ) -> Dict[str, Any]:
+        """Generate diff using the cromper service."""
+        # Encode elf object as base64
+        import base64
```
mkst, line 159: already imported

cromper/Dockerfile
```diff
@@ -0,0 +1,191 @@
...
+COPY --chown=cromper:cromper cromper/ ./cromper/
+COPY --chown=cromper:cromper tests/ ./tests/
+COPY --chown=cromper:cromper conftest.py ./
+COPY --chown=cromper:cromper .env* ./
+
+# Switch to cromper user
+USER cromper
+
+# Install dependencies
+RUN uv sync --frozen
```
mkst, line 143: I don't think we'd do this here, think we'd do it as part of the dev entrypoint.sh

cromper/Dockerfile.prod
```diff
@@ -0,0 +1,189 @@
...
+    && tar xvzf omftools.tar.gz -C /usr/bin jwasm \
+    && rm omftools.tar.gz \
+    && wget "https://github.com/decompals/binutils-omf/releases/download/v0.4/omftools-linux-x86_64.tar.gz" \
+    && tar xvzf omftools-linux-x86_64.tar.gz -C /usr/bin omf-nm omf-objdump \
+    && rm omftools-linux-x86_64.tar.gz
+
+RUN mkdir -p /etc/fonts
+
+ENV WINEPREFIX=/tmp/wine
+
```
mkst, line 123: if we are using the `cromper` user we should use it, if we are using the `ubuntu` user, we should use that (probably my preference given it's UID = 1000) - we shouldn't mix 2 users, adds unnecessary complexity

ethteck:
- [ ] Make compiler, library, and platform endpoints singular
- [ ] Update nginx config
- [ ] Deserialize back into proper objects in Django

ethteck:
@copilot resolve the merge conflicts in this pull request

PR #1673, opened 2025-09-22
Migrate from poetry/black to uv/ruff

ethteck:


.github/workflows/ci.yml
```diff
@@ -42,14 +42,14 @@ jobs:
             -e USE_SANDBOX_JAIL=on \
             -e SANDBOX_DISABLE_PROC=true \
             -e TIMEOUT_SCALE_FACTOR=10 \
-            -e POETRY_VIRTUALENVS_PATH=/backend/virtualenvs \
+
```
mkst, line 45: this newline is breaking the command

mkst:
also need to update the github chore that updates asm-differ/m2c via poetry

PR #1680, opened 2025-09-25
Some progress..

mkst:
Still a long way to go and lot of things are broken. I was able to compile an existing scratch using the GUI with these changes.

```
docker compose up --build
```

after running the python scripts to download compilers/libraries.

Wanted to push this as I've touched a lot of code, and don't want to diverge.

I'm thinking we should have a Compiler and Platform class which have a `.from_dict()` static method to instantiate them - i think that's something I did when I attempted this a few years ago?


ethteck:
> I'm thinking we should have a Compiler and Platform class which have a .from_dict() static method to instantiate them
yeah totally agreed. one of the steps I hadn't gotten to yet

PR #1682, opened 2025-09-26
couple of tweaks

mkst:
really basic handling for `/platform/${id}` and `/compiler/${platform_id}` similar to django.


backend/coreapp/models/scratch.py
ethteck, line 1: I don't think this change is right - it fails the typechecking

PR #1690, opened 2025-09-29
Cromper mkst

mkst:


cromper/tests/test_compilation.py
```diff
@@ -4,7 +4,7 @@
 
 import unittest
 
-from cromper.compiler_wrapper import CompilationResult
+from cromper.cromper.wrappers.compiler_wrapper import CompilationResult
```
ethteck, line 7: can't these be just "cromper.wrappers" and so on? the double cromper is kinda ugly

frontend/src/lib/api/request.ts
```diff
@@ -42,7 +45,10 @@ export class RequestFailedError extends Error {
 }
 
 export function normalizeUrl(url: string) {
-    if (url.startsWith("/")) {
+    // FIXME: handle this properly
+    if (url.startsWith("/platform")) {
```
ethteck, line 49: I think I am missing something, but why do the nginx changes not cover this?

cromper/tests/test_compilation.py
mkst: Urgh this was vscode automatically adjusting the import when I moved the files, will change

PR #1691, opened 2025-10-07
Add more dreamcast compilers

sozud:
Relies on https://github.com/decompme/compilers/issues/44
Seems to work fine locally by manually copying the compilers into backend/compilers/dreamcast


sozud:
I see the CI was kicked off again for this, are the failing tests what is blocking this from being merged? From what I see a bunch of unrelated compilers are failing and it's the same for every PR

PR #1701, opened 2025-10-16
Add Xbox 360 platform

encounter:
~~Currently blocked on https://github.com/decompme/compilers/pull/47 and https://github.com/decompme/decomp.me/pull/1702.~~

~~Without a working `as` or `objdump`, I was planning on updating the frontend to use the new `has_assembler` flag and disable target asm input. We could either display "Manual scratch creation unsupported" or show a file upload input to upload an obj instead.~~

~~For diffing, I was planning on updating the main diff view to force-enable objdiff and use that as the default differ if `!has_objdump`.~~

We now have https://github.com/encounter/xbox360-binutils, so we should be good to go.


mkst:
I think the icon might need some svg massaging

image

and

image

Also, might be nice to tweak the text justification to avoid this:
image

but, other than that, amazing stuff!

image

PR #1779, opened 2025-12-23
Remove exposed ports from dev docker compose config

encounter:
This removes unnecessarily exposed ports from the dev docker-compose.yml. Since the application communicates with these services using the internal docker network, they aren't necessary to be exposed on the host. This avoids port conflicts with frequently-used ports. Additionally, this makes the nginx service bind to port `8080`, which matches the port you'd use to access the site if using regular `yarn dev`.

For compatibility with non-Docker container engines, explicitly specifies `docker.io/library/` prefix for container image refs.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,11 +1,9 @@
 services:
   postgres:
-    image: postgres:17
+    image: docker.io/library/postgres:17
     environment:
       POSTGRES_USER: decompme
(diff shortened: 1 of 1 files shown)


mkst:
.. but I guess commenting them out (rather than removing entirely) would be a happy medium? 

Could you update the DOCKER.md to reflect the nginx port change

PR #1812, opened 2026-02-07
Add instructions for restoring a prod db dump locally

mkst:
I did this earlier and thought it made sense to put the instructions in a doc and share the knowledge.


ethteck:
thanks for writing this! super helpful

PR #1813, opened 2026-02-08
Add GCC 4.4.3 to Android NDK r8e

aloaf812:
I've already added the compiler itself in https://github.com/decompme/compilers/pull/55, this just adds it to the actual website.


mkst:
ok the issue is the tar.bz2 file has an extra folder in it.. so im gonna go ahead and reupload it and regenerate the build.

Issue #1744, opened 2025-11-17
Full search results page


katlyn-edwards:
**Is your feature request related to a problem? Please describe.**
If I'm searching for something that requires more than 5 results, I have to issue the query manually, and parse the Django response. 

**Describe the solution you'd like**
I'd like a full search results page so I can look at all the results, and page through them if there are many.

**Describe alternatives you've considered**
n/a this is common

PR #1854, opened 2026-03-01
Add way to link directly to a search

hollow-owl:
Now search query appears in url so queries can be bookmarked. 

`https://decomp.me/?search=<search query>`

### DIFF
--- a/frontend/src/components/Nav/Search.tsx
+++ b/frontend/src/components/Nav/Search.tsx
@@ -1,6 +1,6 @@
 import { useEffect, useRef, useState } from "react";
 
-import { useRouter } from "next/navigation";
+import { useRouter, useSearchParams } from "next/navigation";
 
 import { SearchIcon } from "@primer/octicons-react";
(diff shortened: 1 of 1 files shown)


mkst:
Not sure what the point of it is though? What problem is being solved? Maybe if we had a dedicated search page it would make sense, but we don't - if you were interested in adding that, that could be cool.

mkst:
> A full search page would be cool.
Yeh we have [an open issue](https://github.com/decompme/decomp.me/issues/1744) for that.

I dont think this PR is particularly nice UX to open the search bar (especially on mobile):
image

For your use case maybe query the API e.g. `https://decomp.me/api/search?name=foo` ? .. but it's still not great (it's still just the results used for the search results) -- I still think this is best solved with a specific results page.

ethteck:
I agree with @mkst here. I like the idea of this PR - linkable search results. But the current search dropdown isn't what should be linked to. We should first add the search page and then allow linking to that.

PR #1898, opened 2026-04-06
Allow --adjust-vma=1234 in diff_flags

mkst:
This scratch has `["--adjust-vma=4096"]` as diff_flags:

image


mkst:
This also opens the door to allowing `--disassemble=` to be modified from the frontend too... Which might be worth doing in the short term (assuming the longer-term is to make some changes to asm-differ)

PR #1902, opened 2026-04-30
Add ido4.1

LLONSIT:
## Important info

- Debug flags are not supported yet
- The compile command uses a tool named [`ecoff_tool`](https://gist.github.com/Mc-muffin/74bc79cb84988cf4ad3b4bc64c58d3be) (made by @Mc-muffin) to convert the output ECOFF object in to an ELF file

### DIFF
--- a/backend/coreapp/compilers.py
+++ b/backend/coreapp/compilers.py
@@ -855,7 +855,17 @@ MWCCPSP_3_0_1_219 = MWCCPSPCompiler(
     cc=MWCCPSP_CC,
 )
 
+MWCCEPPC_CC = 'printf "%s" "${COMPILER_FLAGS}" | xargs -x -- ${WIBO} "${COMPILER_DIR}/mwcceppc.exe" -pragma "msg_show_realref off" -c -proc gekko -nostdinc -stderr -o "${OUTPUT}" "${INPUT}"'
+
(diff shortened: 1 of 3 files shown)


mkst:
You'll need to add it to the backend/compilers/download.py too 👍

backend/coreapp/compilers.py
```diff
@@ -855,7 +855,15 @@ def available_platforms() -> List[Platform]:
     cc=MWCCPSP_CC,
 )
 
+IDO_41_CC = '"${COMPILER_DIR}"/usr/bin/qemu-irix-4.0 -silent -L "${COMPILER_DIR}" "${COMPILER_DIR}/usr/bin/cc" -I "{COMPILER_DIR}"/usr/include -EL -c -Xcpluscomm -G0 -non_shared ${COMPILER_FLAGS} -o "${OUTPUT}" "${INPUT}"'
```
mkst, line 858: `-c` is fine but im wondering if the user should be able to specify `-Xcpluscomm -G0 -non_shared`? although I guess if a user *does* add more flags these will override the defaults set here?

LLONSIT: > if a user does add more flags these will override the defaults set here?

Yes

mkst:
Your ido4.1 archive has the folder inside :( I will re-upload it and update the compilers repo.

```
ubuntu@a788901cd63b:/backend/compilers/n64$ ll ido4.1/ido4.1/
total 16
drwxr-xr-x 4 ubuntu ubuntu 4096 May  1 07:30 ./
drwxr-xr-x 3 ubuntu ubuntu 4096 May  5 18:37 ../
drwxr-xr-x 2 ubuntu ubuntu 4096 May  1 07:30 lib/
drwx------ 5 ubuntu ubuntu 4096 May  1 07:30 usr/
```

PR #1910, opened 2026-05-04
Codex refactors

mkst:
Mostly pulling out helper functions to allow them to be tested. Also fixes a small bug where navigating to the Compiler Options tab triggers a re-compile request.

Blurb from codex:
> This PR extracts several pieces of scratch/editor state logic into small pure helpers with unit coverage: new scratch draft transitions, compiler/diff flag editing, parameter validation, library editing, and scratch save/compile request payloads. It also tightens a few no-op paths so mounting or interacting with compiler options does not mark a scratch dirty or trigger recompilation when nothing actually changed. Frontend unit tests now cover those extracted behaviours, alongside the existing lint/type checks.


backend/coreapp/diff_wrapper.py
```diff
@@ -3,25 +3,33 @@
...
+from .wrapper_result import DiffResult
 
 logger = logging.getLogger(__name__)
 
 MAX_FUNC_SIZE_LINES = 25000
 
+if TYPE_CHECKING:
+    F = TypeVar("F")
+
+    def lru_cache(maxsize: int = 128, typed: bool = False) -> Callable[[F], F]:
```
bates64, line 27: why can't you just import lru_cache unconditionally? it's not like there's a circular dependency
or is it because their types are incorrect or something

mkst: Same behaviour as in compiler_wrapper: https://github.com/decompme/decomp.me/blob/d66311dfd15ab864274b7084554015f7a640b2fe/backend/coreapp/compiler_wrapper.py#L29

PR #1919, opened 2026-05-09
Upgrade next from 15 to 16

mkst:


mkst:
Vercel failed due to

> The Edge Function "opengraph-image" size is 1.04 MB and your plan size limit is 1 MB.

so maybe we just go with the default of fonts instead...

PR #1943, opened 2026-05-19
Add 'delete scratch' button to user profile page

MatBourgon:
These changes add a 'delete scratch' button to the user's profile, only viewable and usable by themselves and admins.
The button starts white and prompts for confirmation before deletion, but can be bypassed by holding shift (which also turns the button red).

image
image

### DIFF
--- a/frontend/src/components/ScratchItem.module.scss
+++ b/frontend/src/components/ScratchItem.module.scss
@@ -147,3 +147,11 @@
         text-overflow: ellipsis;
     }
 }
+
+.red-on-shift {
+    color: rgb(206, 58, 58);
+}
+
+.red-on-shift:hover {
+    color: rgb(245, 141, 141) !important;
+}
--- a/frontend/src/components/ScratchItem.module.scss
+++ b/frontend/src/components/ScratchItem.module.scss
@@ -150,14 +150,18 @@
 
 .delete-button {
     --warning-color: #ce3a3a;
-    color: var(--warning-color)  !important;
+    background-color: var(--warning-color)  !important;
     border-color: var(--warning-color) !important;
+    color: white !important;
     font-size: 0.8em;
     padding: 0 0.4em;
+    min-width: 4.5rem;
+    justify-content: center;
 }
 
 .delete-button:hover {
     --warning-hover-color: #f58d8d;
-    color: var(--warning-hover-color)  !important;
+    background-color: var(--warning-hover-color)  !important;
     border-color: var(--warning-hover-color) !important;
+    color: white !important;
 }
--- a/frontend/src/components/ScratchItem.module.scss
+++ b/frontend/src/components/ScratchItem.module.scss
@@ -149,9 +149,13 @@
 }
 
 .red-on-shift {
-    color: rgb(206, 58, 58);
+    --warning-color: rgb(206, 58, 58);
+    color: var(--warning-color)  !important;
+    border-color: var(--warning-color) !important;
 }
 
 .red-on-shift:hover {
-    color: rgb(245, 141, 141) !important;
+    --warning-hover-color: rgb(245, 141, 141);
+    color: var(--warning-hover-color)  !important;
+    border-color: var(--warning-hover-color) !important;
 }
--- a/frontend/src/components/ScratchItem.module.scss
+++ b/frontend/src/components/ScratchItem.module.scss
@@ -148,13 +148,15 @@
     }
 }
 
-.red-on-shift {
+.delete-button {
     --warning-color: #ce3a3a;
     color: var(--warning-color)  !important;
     border-color: var(--warning-color) !important;
+    font-size: 0.8em;
+    padding: 0 0.4em;
 }
 
-.red-on-shift:hover {
+.delete-button:hover {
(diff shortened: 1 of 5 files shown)


frontend/src/components/ScratchItem.tsx
```diff
@@ -144,53 +145,120 @@ function ScratchItemRow({
...
-                    <ScratchMetadata
-                        scratch={scratch}
-                        showPresetOrCompiler={showPresetOrCompiler}
-                    />
-                    {(children || showOwner) && (
-                        <div className={styles.metadataAside}>
-                            {children && (
-                                <div className={styles.actions}>{children}</div>
+        <>
+            {" "}
```
mkst: is this whitespace necessary (also at the end)?

mkst:
I'm wondering if the delete button should be a filled red pill that's like `[🗑️ delete]` to match the 100% pill.. perhaps the pressing shift doesnt need to change the visualisation, just suppresses the prompt?

frontend/src/components/ScratchItem.tsx
MatBourgon: The linter had added those in, so I had assumed removing them would have done nothing; I should have at least checked. Looks like I can remove them.

MatBourgon:
I've gone ahead and done that, so it no longer reacts to pressing shift, but the confirmation still gets skipped. I've also resized the button to match the pill's size a bit better:
image

Unless I misunderstood the comment and by 'filled' you meant red bg with white text, in which case I can quickly fix as well.

mkst:
Yeh I mean so it looks like it belongs next to the 100% pill 😎

MatBourgon:
Sounds good 😎
Here's how it looks like now:
image

mkst:
Nice!

Could you remove the `translate y` that i added to the progress pill css... When i was looking on my phone i needed to shift the text down, but now every time i look at it, its definitely wrong to have done so... 

https://github.com/decompme/decomp.me/blob/2d0becd9f8776430e2bd52cd87dbdc5cacfe3ca6/frontend/src/components/ScratchItem.module.scss#L53 
This ^^ just change the 1px to 0, and/or make it translateX instead

mkst:
What would actually be pretty cool, is to hide the "Delete" word on mobile, so its just the trash can icon on small screens. 

You can do something similar to this code : 
https://github.com/decompme/decomp.me/blob/2d0becd9f8776430e2bd52cd87dbdc5cacfe3ca6/frontend/src/components/Scratch/ScratchToolbar.tsx#L175

MatBourgon:
I've handled the fix to the translation, but I'm not too sure about the mobile styling. It doesn't quite look right.
image

Unless you're fine with this.

mkst:
ah no i meant to shrink the width to be more like a square ignore my poor paint skills but like this (but with the trash icon):

image

MatBourgon:
Oh gotcha. I'm not too sure how to apply custom stylings specifically for mobile so I can't quite get it to be square shaped, but I've got it like this now:

image
image

frontend/src/components/ScratchItem.module.scss
```diff
@@ -147,3 +146,21 @@
         text-overflow: ellipsis;
     }
 }
+
+.delete-button {
+    --warning-color: #ce3a3a;
```
mkst: it might be worth trying to use tailwind for styling .. we're slowly moving away from the explicit (s)css to using tailwind. might help with all the `!important`s

frontend/src/components/ScratchItem.tsx
```diff
@@ -144,53 +145,115 @@ function ScratchItemRow({
...
+            !confirm(
+                "Are you sure you want to delete this scratch? This action cannot be undone.",
+            )
+        ) {
+            return;
+        }
+
+        try {
+            await api.delete_(scratchUrl(scratch), {});
+            setShowElement(false); // Hide deleted element to avoid performing a page refresh, and allow deleting more scratches
```
mkst: maybe we should disable the delete (have some `const [isDeleting, setIsDeleting] = useState(false)` and then set isDeleting before the delete (and disable the button to avoid users double-clicking?

frontend/src/components/ScratchItem.module.scss
MatBourgon: I gave it a shot, but I'm pretty unfamiliar with Tailwind and the styling wasn't been shown. I'm not sure if I need to run a command to build anything, or if the button was overriding the style anyway.

If this is a blocker, then someone else may need to implement it, or if I know how to properly get it to show the UI changes I can give it another shot, but as it currently stands I cannot do it.

mkst: Let me take a look tonight and I'll PR something into your branch if I get anywhere 😊

MatBourgon:
Just reposting the images here so they're tracked in this PR. Thanks for the help!

Desktop:
image

Mobile:
image

mkst:
Nice - sorry for the biome failure, are you good to fix that?

MatBourgon:
Should be a quick fix yeah.

mkst:
LGTM :rocket:

PR #1952, opened 2026-05-28
Reduce ulimit before running dosemu to fix performance regression

mkst:
See https://github.com/dosemu2/dosemu2/issues/2889 for some background. dosemu2 tries to close all file descriptors.. which is equal to the ulimit (1048576) which takes AGES. reducing this down to even 16k means invocation goes from 1.3 seconds to 80ms and should not affect compilation.

### DIFF
--- a/backend/coreapp/compilers.py
+++ b/backend/coreapp/compilers.py
@@ -387,6 +387,8 @@ CLANG_800 = ClangCompiler(
 )
 
 # PS1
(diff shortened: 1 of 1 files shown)


mkst:
Gah, this should have fixed the timeouts :(

mkst:
dosemu devs have fixed this issue so this change is no longer required... Just need to wait for their release cycle to complete.

PR #1953, opened 2026-05-29
Run tests with DEBUG=1 in CI

mkst:


mkst:
Need to use `--debug-mode` instead. Anyhow gonna close this one too.

PR #1962, opened 2026-06-04
Add Microsoft C Compiler 5.1

mkst:


SilverCard:
Hello, and thanks for this!

I'm looking at the line below of this PR.

"'(HOME="$(pwd)" LD_LIBRARY_PATH=/usr/lib/x86_64-linux-gnu/i386-pc-dj64/lib64 /usr/bin/dosemu -quiet -dumb -f .dosemurc -p -K . -E "D:\\BIN\\CL.EXE /c **/AM** ${COMPILER_FLAGS} /Foout.o dos_src.c") && '"

I think that the compiler flag "/AM" should be removed, since it flags the compiler to use the "medium memory model". It should be left to the user to use this flag if he needs it.

mkst:
I was going off the COMPILE.BAT in the zip... what impact does it have on codegen?

SilverCard:
Sorry, i left it by mistake, i was testing it. 

It does have a lot of impact on codegen, since it changes the pointer type of the dos code between far and near. I'm not an expert on this, but i think to have a match, the correct memory model should be set via the compiler flag.

I was testing it on the medium memory model, so i left the /AM flag by mistake.

image

mkst:
Cool, will fix it :)

PR #1979, opened 2026-06-11
Add gcc2.96

SBird1337:
Adds backend boilerplate for GCC 2.96, depends on https://github.com/decompme/compilers/pull/73 .

GCC 2.96 is a development snapshot between GCC 2.95 "agbcc" and GCC 3 which already has a unified thumb/arm compiler in one binary (`-mthumb`, `-marm`).

It was used in the game Golden Sun for the GBA.

If possible it would be nice to have a compiler flag preset for Golden Sun 1, it uses caller-save r4 to perform some arcane magic:
```
-O2 -mthumb -mthumb-interwork -mcpu=arm7tdmi -fcall-used-r4
```

### DIFF
--- a/backend/coreapp/compilers.py
+++ b/backend/coreapp/compilers.py
@@ -296,9 +296,9 @@ OLD_AGBCC = GCCCompiler(
 )
 
 GCC_296 = GCCCompiler(
-    id = "gcc2.96",
+    id="gcc2.96",
     platform=GBA,
-    cc='"${COMPILER_DIR}"/usr/local/bin/arm-elf-gcc $COMPILER_FLAGS -S -o - "$INPUT" | arm-none-eabi-as -mcpu=arm7tdmi -o "$OUTPUT"'
+    cc='"${COMPILER_DIR}"/usr/local/bin/arm-elf-gcc $COMPILER_FLAGS -S -o - "$INPUT" | arm-none-eabi-as -mcpu=arm7tdmi -o "$OUTPUT"',
 )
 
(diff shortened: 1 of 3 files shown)


backend/coreapp/compilers.py
```diff
@@ -295,6 +295,12 @@ def available_platforms() -> list[Platform]:
     base_compiler=AGBCC,
 )
 
+GCC_296 = GCCCompiler(
+    id = "gcc2.96",
+    platform=GBA,
+    cc='"${COMPILER_DIR}"/usr/local/bin/arm-elf-gcc $COMPILER_FLAGS -S -o - "$INPUT" | arm-none-eabi-as -mcpu=arm7tdmi -o "$OUTPUT"'
```
mkst: any reason you aren't using `usr/local/bin/arm-elf-as` ? wont this affect matching?

SBird1337: I don't think it should, most GBA games use modern gnu `as` as the assembler. We certainly do for the gs-decomp project as building the whole toolchain is a little more involved than just getting gcc / cc1.

E: To clearify: I kinda just mirrored a) the other gba compilers, b) our own Makefile :)

mkst: If you guys are getting matches without needing era-specific assembler then all good - was just flagging to make sure it wasnt a mistake :)

mkst:
Created a preset: https://decomp.me/preset/236. I can assign you as the owner once you've created a scratch on the site.

PR #1996, opened 2026-06-17
Fix Watcom 16-bit codegen compiler selection

### LINKED ISSUE #1645: [BUG] Watcom invalid option with "-0"
**Describe the bug**
Trying to use either `-0`, `-1` or `-2` (Optimize for i286/i186/i086) with any Watcom compiler fails to compile with the error "E1073: Invalid option: '-0'"

I believe this is because `WATCOM_CC` is always `wcc386` [here](https://github.com/decompme/decomp.me/blob/2fe4e457719f9dc13d313b33445c34429620bdb8/backend/coreapp/compilers.py#L1409) which according to its help output only supports `-3r` and above. For these older targets `wcc` should be used which uses optimization settings `-0` -> `-6`

**To Reproduce**
1. Go to https://decomp.me/scratch/3iFaP
2. Click on Options
3. Change "Code generation" to "-0"
4. Expand Problems panel to see error

**Expected behavior**
Compilation without error

nkgotcode:
Watcom C compiler presets using 16-bit code-generation flags now invoke `wcc.exe` instead of `wcc386.exe`, while 386 C and C++ presets keep their existing compiler binaries.

Fixes #1645

Validation:

```bash
uv run python manage.py test coreapp.tests.test_compilation.CompilationTests.test_watcom_16_bit_codegen_uses_16_bit_compiler coreapp.tests.test_compilation.CompilationTests.test_watcom_32_bit_codegen_uses_386_compiler coreapp.tests.test_compilation.CompilationTests.test_watcom_cpp_keeps_386_compiler -v 2
uv run ruff check coreapp/compiler_wrapper.py coreapp/tests/test_compilation.py
uv run ruff format --check coreapp/compiler_wrapper.py coreapp/tests/test_compilation.py
```

All commands passed.

### DIFF
--- a/backend/coreapp/compiler_wrapper.py
+++ b/backend/coreapp/compiler_wrapper.py
@@ -43,6 +43,23 @@ def _check_assembly_cache(*args: str) -> tuple[Assembly | None, str]:
 
 class CompilerWrapper:
     @staticmethod
+    def get_compiler_command(compiler: Compiler, compiler_flags: str) -> str:
+        cc_cmd = compiler.cc
+
+        if (
+            isinstance(compiler, compilers.WatcomCompiler)
+            and compiler.language == Language.C
+            and any(flag in {"-0", "-1", "-2"} for flag in compiler_flags.split())
+        ):
+            cc_cmd = cc_cmd.replace("/wcc386.exe", "/wcc.exe")
+
+        # IDO hack to support -KPIC
+        if compiler.type == CompilerType.IDO and "-KPIC" in compiler_flags:
+            cc_cmd = cc_cmd.replace("-non_shared", "")
+
+        return cc_cmd
+
+    @staticmethod
     def filter_compiler_flags(compiler_flags: str) -> str:
         # Remove irrelevant flags that are part of the base compiler configs or
         # don't affect matching, but clutter the compiler settings field.
@@ -128,7 +145,7 @@ class CompilerWrapper:
                 f.write(code)
                 f.write("\n")
 
-            cc_cmd = compiler.cc
+            cc_cmd = CompilerWrapper.get_compiler_command(compiler, compiler_flags)
 
             # MWCC requires the file to exist for DWARF line numbers,
             # and requires the file contents for error messages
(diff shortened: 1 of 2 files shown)


mkst:
I'm not sure about this approach, but if we are going to have dynamic compiler commands, i think it should belong to the Compiler class rather than the CompilerWrapper class.

My preference would probably be adding a custom driver (i.e. a GCC-like utility) that invokes the correct watcom compiler based on the flags passed to it, rather than this logic living inside decomp.me

nkgotcode:
Thanks, that makes sense. For this PR, would a small rework that moves the dynamic command selection onto the `Compiler` side be acceptable, or would you rather close this and leave the Watcom flag dispatch for a separate custom driver/tooling change?

mkst:
The *usual* approach we take has been just add another compiler -> we have C vs C++ versions of quite a few flavours... but we already have 2 flavours of each of the Watcom compilers (for wcc386.exe vs wpp386.exe) so I'm not sure it would be great to do that *again* here. Is there no driver that picks the "right" watcom flavour?

nkgotcode:
In the current decomp.me config I only found direct `wcc386.exe` / `wpp386.exe` invocations in `backend/coreapp/compilers.py`; I do not see a configured Watcom selector driver in the repo. The bug report points to `wcc` for `-0`/`-1`/`-2`, but if there is a driver in the compiler bundle you would prefer to use, I can rework this toward that instead.

mkst:
Been chatting to my friend ChatGPT.. would using `wcl.exe` (and/or wcl386.exe) work?

nkgotcode:
I checked this against the Watcom 10.5 bundle: `binnt/wcl.exe` is not present, and `binnt/wcl386.exe` still rejects the older codegen flags (`-0`, `-1`, `-2`) with `E1073: Invalid option`; `wcl386.exe -3r` does create an object file. So `wcl386.exe` does not appear to solve this case.

mkst:
Out of interest, are you a user of the site - are you raising this PR because it solves a problem you're having? or simply because there's an issue raised?

nkgotcode:
I'm not a regular user of the site; I picked this up because the open issue was reproducible and the fix looked small. If you'd prefer not to carry this app-side, I can close the PR.

mkst:
Gonna close this, im not convinced the wcc.exe actually works.

PR #2001, opened 2026-06-19
Github link in FAQ page is invalid

Ed-1T:
The link to this repo from the FAQ page points to
`https://github.com/decomp.me`
which is invalid and doesn't redirect to this repository

The correct URL should therefore either be
`https://github.com/decompme` (organization page)
or
`https://github.com/decompme/decomp.me` (repository page)

I'm assuming in that context it was meant to redirect to the repository,
so the proposed changes will reflect that


mkst:
Good catch, thank you!

PR #2044, opened 2026-07-19
Turn on m2c for saturn

sozud:
This enables m2c for saturn and adds a unit test. I tested locally and it seemed to work fine.


sozud:
mypy failure looks like an ephemeral github actions issue `Error: fetch failed`

PR #2061, opened 2026-08-01
add best functions by name page

sozud:
This puts together a "best scratches by function name page. The use case is:

On the projects I work on, people tend to get a function to 90%+ then get stuck, and people forget about the scratch. I want to be able to go back and see these scratches. My function_finder tool used to work for this case by using the API but now with CloudFlare it's non-functional. https://github.com/Xeeynamo/sotn-decomp/blob/master/tools/function_finder/function_finder_psx.py

This adds:

`GET /api/scratch/best-by-name` groups scratches by function name, ranks by match %, filterable by platform/preset/depth/min-match/search

`/scratches/best?platform=saturn&preset=<id>` page consuming the api

Screenshot with fake data:

image

### DIFF
--- a/backend/coreapp/views/scratch.py
+++ b/backend/coreapp/views/scratch.py
@@ -688,9 +688,7 @@ class ScratchViewSet(
     @action(detail=False, methods=["GET"], url_path="best-by-name")
     def best_by_name(self, request: Request) -> Response:
         if not request.query_params.get("preset"):
-            raise serializers.ValidationError(
-                {"detail": "'preset' is required."}
-            )
+            raise serializers.ValidationError({"detail": "'preset' is required."})
 
         try:
             depth = int(request.query_params.get("depth", DEFAULT_BEST_BY_NAME_DEPTH))
--- a/backend/coreapp/views/scratch.py
+++ b/backend/coreapp/views/scratch.py
@@ -687,11 +687,9 @@ class ScratchViewSet(
     @method_decorator(globally_cacheable(max_age=5, stale_while_revalidate=1))
     @action(detail=False, methods=["GET"], url_path="best-by-name")
     def best_by_name(self, request: Request) -> Response:
-        if not request.query_params.get("platform") or not request.query_params.get(
-            "preset"
-        ):
+        if not request.query_params.get("preset"):
             raise serializers.ValidationError(
-                {"detail": "Both 'platform' and 'preset' are required."}
+                {"detail": "'preset' is required."}
             )
 
         try:
--- a/backend/coreapp/views/scratch.py
+++ b/backend/coreapp/views/scratch.py
@@ -3,15 +3,17 @@ import hashlib
 import io
 import json
 import logging
+import math
 import re
 import zipfile
+from collections import defaultdict
 from datetime import datetime
-from typing import Any
+from typing import Any, cast
 
 import django_filters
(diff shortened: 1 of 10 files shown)


mkst:
Need to look into this, but feel like it's potentially an expensive query (if it's not based on family) that perhaps could be something we do based on a dump of the database, or some daily batch process or something?

sozud:
It forces selecting a platform and preset to avoid a full scan. I ran EXPLAIN ANALYZE and saw it used the preset_id index. On my system (old 2020 macbook) it was about 5ms to do the three queries on 398 rows.

I'm not sure what preset has the most scratches but I see Rhythm Tengoku has more than average at 4241. I tested that and it took about 15ms. I could turn up the caching if it seems worthwhile.

backend/coreapp/views/scratch.py
```diff
@@ -630,3 +646,128 @@ def family(self, request: Request, pk: str) -> Response:
...
+                    {"min_match": "Must be a number between 0.0 and 1.0."}
+                )
+            queryset = queryset.filter(match_percent__gte=min_match_value)
+
+        return queryset
+
+    @method_decorator(globally_cacheable(max_age=5, stale_while_revalidate=1))
+    @action(detail=False, methods=["GET"], url_path="best-by-name")
+    def best_by_name(self, request: Request) -> Response:
+        if not request.query_params.get("platform") or not request.query_params.get(
```
mkst: preset dictates platform though, why are both needed?

backend/coreapp/views/scratch.py
```diff
@@ -630,3 +646,128 @@ def family(self, request: Request, pk: str) -> Response:
         return Response(
             TerseScratchSerializer(family, many=True, context={"request": request}).data
         )
+
+    def _best_by_name_eligible_queryset(self, request: Request) -> QuerySet[Scratch]:
+        queryset = (
+            Scratch.objects.all()
+            .exclude(name=UNTITLED_SCRATCH_NAME)
```
mkst, line 653: why? doesnt this just add an extra clause to the search (i.e. its more unhelpful than helpful)?

mkst:
I'm still not super keen on the general concept of this functionality.. but will bow to other maintainers opinion. I feel like we should be able to fix your cloudflare issue easily enough (we have other known users using the api -> reach out to us on the discord).

backend/coreapp/views/scratch.py
sozud: My original reasoning is that the Untitleds are random unrelated scratches so getting a ranking of % complete doesn't make sense. I checked this with a synthetic dataset of 100,000 scratches locally with 10% being Untitleds and I got:
216ms with Untitled exclusion
231ms without Untitled exclusion
I'm not sure what the real ratio is like on the site

backend/coreapp/views/scratch.py
sozud: You're right, revised this to remove platform

sozud:
> I feel like we should be able to fix your cloudflare issue easily enough (we have other known users using the api -> reach out to us on the discord).

I most recently asked about this on August 1 but didn't get a response https://discord.com/channels/897066363951128586/897075447534354462/1532938403036004665
My reasoning is that this provides a way to get this information without having to request access to the API or having to adjust to the abuse prevention changes as they evolve

PR #2075, opened 2026-08-18
Add ido7.1_irix4_c++

jtl3d:
This I think may be the solution to SSB64's ovl8 issues. It is C++ with IRIX4, and in my local testing I am able to get things matching without resorting to weird file splits.

### DIFF
--- a/backend/coreapp/compilers.py
+++ b/backend/coreapp/compilers.py
@@ -941,7 +941,7 @@ IDO71_IRIX4_CXX = IDOCompiler(
     platform=N64,
     base_compiler=IDO71,
     language=Language.OLD_CXX,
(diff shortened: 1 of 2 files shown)


jtl3d:
Uses the same script as the C IRIX4 one.

jtl3d:
I don't think the automated check errors are related to my code

mkst:
I think they are?

ethteck:
yeah the test that is failing is the compiler test for the new compiler

jtl3d:
Oh. Do I need to update the test then to not have it fail on the bad context hardcoded in the test?

jtl3d:
Nevermind figured out the problem, tests passing locally for me now. Updated compilers PR: https://github.com/decompme/compilers/pull/81

PR #2088, opened 2026-09-01
Redo the way flags are serialized

ethteck:


backend/coreapp/flags.py
```diff
@@ -184,12 +230,11 @@ def to_json(self) -> dict[str, str]:
     Checkbox(id="shc_inline", flag="-inline"),
 ]
 
-COMMON_SHC_FLAGS: Flags = COMMON_SHC_OLD_FLAGS + [
+SHC_FLAGS: Flags = [
```
mkst: Something feels a bit fishy here. because we're creating global flags to create `COMPILER_FLAG_CLASSES`, but then we only refer to them by string name (e.g. `diff_flag_class="mips",`).

Should these all be FlagClasses instead of Flags? or should the diff_flag_class use like `COMPILER_FLAG_CLASSES["mips"]` ? I dunno, feels a little fragile / or maybe feels like it could be simplified?

frontend/next-env.d.ts
```diff
@@ -1,6 +1,6 @@
 /// <reference types="next" />
 /// <reference types="next/image-types/global" />
-import "./.next/types/routes.d.ts";
+import "./.next/dev/types/routes.d.ts";
```
mkst: dont check this in

backend/coreapp/flags.py
ethteck: yeah.. I agree. how about we have the flags be a dataclass and it stores its class name inside, so instead of referring to raw strings we do `COMON_SHC_OLD_FLAGS.name` ?

backend/coreapp/compilers.py
```diff
@@ -10,30 +10,7 @@
...
-    COMMON_MSVC_FLAGS,
-    COMMON_MWCC_NDS_ARM9_FLAGS,
-    COMMON_MWCC_PS2_FLAGS,
-    COMMON_MWCC_PSP_FLAGS,
-    COMMON_MWCC_WII_GC_FLAGS,
-    COMMON_SHC_FLAGS,
-    COMMON_SHC_OLD_FLAGS,
-    COMMON_WATCOM_FLAGS,
-    Flags,
-    Language,
```
mkst, line 35: did Language get removed from flags.py?

mkst:
- Seems we've added the compiler `id` to the dictionary? is that necessary if it's already the key?
- perhaps `class` should be `flags_class` (and `diff_class` should be `diff_flags_class` ?)
  - maybe it makes sense as class/flags_class if `class` will mean more than just flags?
-  if we're adding `id` then perhaps we could/should return list of compilers rather than a dict, but obviously that has implications for e.g. permuter and anyone else expecting a dictionary.

**before**
```json
{
    "compilers": {
        "agbcc": {
            "platform": "gba",
            "flags": [
                {
                    "type": "flagset",
                    "id": "gcc_opt_level",
                    "flags": [
                        "-O0",
                        "-O1",
                        "-O2",
                        "-O3"
                    ]
                },
                {
                    "type": "flagset",
                    "id": "gcc_debug_level",
                    "flags": [
                        "-gdwarf-2",
                        "-gdwarf",
                        "-g0",
                        "-g1",
                        "-g2",
                        "-g3"
                    ]
                },
                {
                    "type": "flagset",
                    "id": "gcc_char_type",
                    "flags": [
                        "-fsigned-char",
                        "-funsigned-char"
                    ]
                },
                {
                    "type": "checkbox",
                    "id": "gcc_force_addr",
                    "flag": "-fforce-addr"
                }
            ],
            "diff_flags": [
                {
                    "type": "flagset",
                    "id": "diff_algorithm",
                    "flags": [
                        "-DIFFlevenshtein",
                        "-DIFFdifflib"
                    ]
                },
                {
                    "type": "checkbox",
                    "id": "diff_function_symbols",
                    "flag": "-DIFFdiff_function_symbols"
                },
                {
                    "type": "parameter",
                    "format": "int-or-hex",
                    "id": "adjust_vma",
                    "flag": "--adjust-vma"
                }
            ]
        },
```
after:
```json
{
    "compilers": {
        "agbcc": {
            "id": "agbcc",
            "platform": "gba",
            "class": "gcc",
            "diff_class": "common"
        },
```

PR #2092, opened 2026-09-04
Prioritize exact scratch name matches during search, allow 3x page_size scratch results

### LINKED ISSUE #2090: Search: Prioritize exact scratch name searches
Currently searches can show so many results that the scratch you are searching for does not appear in the search. This is understandable, however in cases where the one that is not present is one that is an identical match, that should be prioritized ahead of results that match and have extra appended.

For example, searching `tmevReset` came up with many functions that had `tmevReset` prepended to other words

Image

However, hidden from the searches was a result that is an exact match, [`tmevReset`](https://decomp.me/scratch/aH4gF).

I believe this would be a good enhancement for scratch discoverability. I currently track my scratches in a text file so I was able to find it that way instead. But in the case where this happened when searching someone else's scratches it seems one would need to find it via the API call.

___
If a search has many results and some are skipped *and* there is exact query matches for the query string that those are shown ahead of other results.

For the example above, `tmevReset` scratch would be first, and one of the `tmevReset ...` scratches would simply be pushed off of the results list.
___
While this would not be an alternative in the traditional sense because it would still be an issue for those searching for someone else's scratch, having a search bar within one's own scratches page would improve discoverability of one's scratches while not having as big of a search scope as searching everyone's scratches globally for a name.

E.g. if I went to my scratches and see I have over a thousand, instead of scrolling down and clicking "Load More" and Ctrl+F'ing for it, I could search my thousand scratches and see the all results would help assist what may be considered a discoverability challenge.

mkst:
Closes #2090.

Also given we allow 15 results (users+presets+scratches), I thought it makes sense to allow `15 - user results - preset results` scratch results, given that is the most common use of the search (i.e. don't limit to just 5 scratch results).

### DIFF
--- a/backend/coreapp/tests/test_search.py
+++ b/backend/coreapp/tests/test_search.py
@@ -1,10 +1,23 @@
 from django.urls import reverse
 from rest_framework import status
 
+from coreapp.models.scratch import Assembly, Scratch
 from coreapp.tests.common import BaseTestCase
 
 
 class SearchTests(BaseTestCase):
+    def create_search_scratch(self, name: str) -> Scratch:
+        assembly = Assembly.objects.create(
+            hash=f"{name}-assembly",
+            arch="dummy",
+        )
+        return Scratch.objects.create(
+            name=name,
+            compiler="dummy",
+            platform="dummy",
+            target_assembly=assembly,
+        )
+
     def test_rejects_long_query(self) -> None:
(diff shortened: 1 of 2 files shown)


mkst:
For future me: Need to update the libncurses package being pulled down.(https://security.ubuntu.com/ubuntu/pool/universe/n/ncurses)

PR #2094, opened 2026-09-04
Add intel 5.0 C++ compiler

bwrsandman:
- [x] https://github.com/decompme/compilers/pull/86

Adding icc 5 to the list of compilers.

This compiler was used to take advantage of SIMD around 2000. The game I decomp, Black & White, used for one first party file, it linked with libircmt.lib from this compiler and also liked with AMaths (approximate maths) which was also compiled with icc 5.

This compiler would work with msvc 6.0 for the majority of the standard lib, runtime libs and ide.

After searching for years, I found it on archive.org. I also found the linux version on a university website.

Note that this build is not exactly the build used in Black & White. That game used a beta build which is lost to time.


mkst:
Could you rebase off `main` and push? I feel like GitHub isnt running our PRs against latest main for some reason...

